PR #11, opened 2015-07-19
Add plFollowMod.

Deledrius:


korman/properties/modifiers/render.py
```diff
@@ -171,3 +171,57 @@ def export(self, exporter, bo, so):
...
+
+    leader_object = StringProperty(name="Leader Object",
+                                   description="Object to follow")
+
+    def export(self, exporter, bo, so):
+        fm = exporter.mgr.find_create_object(plFollowMod, so=so, name=self.key_name)
+
+        fm.mode = 0
+        for flag in [getattr(plFollowMod, mode) for mode in self.follow_mode]:
+            fm.mode |= flag
```
Hoikas:
```python
if self.follow_mode:
    for flag in (getattr(plFollowMod, mode) for mode in self.follow_mode):
        fm.mode |= flag
else:
    fm.mode = 0
```

Yours iterates twice, mine iterates once :wink:

korman/ui/modifiers/render.py
```diff
@@ -55,3 +55,10 @@ def viewfacemod(modifier, layout, context):
         col = split.column()
         col.enabled = modifier.offset
         col.prop(modifier, "offset_coord")
+
+def followmod(modifier, layout, context):
+    layout.row().prop(modifier, "follow_mode", expand=True)
+    layout.row().prop(modifier, "leader_type")
+    if modifier.leader_type == "kFollowObject":
+        layout.prop_search(modifier, "leader_object", bpy.data, "objects", icon="OUTLINER_OB_MESH")
+        layout.separator()
```
Hoikas: O.o

korman/ui/modifiers/render.py
```diff
@@ -55,3 +55,10 @@ def viewfacemod(modifier, layout, context):
         col = split.column()
         col.enabled = modifier.offset
         col.prop(modifier, "offset_coord")
+
+def followmod(modifier, layout, context):
+    layout.row().prop(modifier, "follow_mode", expand=True)
+    layout.row().prop(modifier, "leader_type")
```
Hoikas: Are the new rows strictly needed?

korman/properties/modifiers/render.py
Deledrius: Yours also will always be bound to X,Y,Z no matter what the selection is, because the default value isn't 0.

Hoikas: check the if ;)

Hoikas: bah, crazy me, you're right about that. I do want to see the generator though

Deledrius: I mean the default value in the plFollowMod, not the UI.

korman/ui/modifiers/render.py
Deledrius: The first one, yes.  The second one, no.  Fixed.

Hoikas: oh, yeah, ugly column enum

PR #23, opened 2016-01-18
Soft Volumes and Vis Regions

### LINKED ISSUE #17: Add Soft Volumes

Hoikas:
This fixes #17 and #18.


Deledrius:
Aside from the envelope in-game behaving as if it's bigger than the actual region I made, this is good.

PR #24, opened 2016-01-20
Fix typo in exception name so error reporting doesn't throw an error.

Deledrius:
This gives a much better error when it works correctly. :smirk:


Hoikas:
:trollface:

PR #28, opened 2016-02-07
Animated physicals + no clamped stencils

Jrius:
Set mass=1 and kPinned for non-kickable physicals having a CoordinateInterface, so they can be animated.
Don't force clamping for texture stencils, so they can be tiled if necessary.

More infos : [https://forum.guildofwriters.org/viewtopic.php?f=2&t=6574&p=69469#p69469](url)

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -166,7 +166,7 @@ class MaterialConverter:
         if slot.use_stencil:
             hsgmat.compFlags |= hsGMaterial.kCompNeedsBlendChannel
             state.blendFlags |= hsGMatState.kBlendAlpha | hsGMatState.kBlendAlphaMult | hsGMatState.kBlendNoTexColor
-            if slot.texture and slot.texture.type == "BLEND":
+            if slot.texture.type == "BLEND":
                 state.clampFlags |= hsGMatState.kClampTexture
             state.ZFlags |= hsGMatState.kZNoZWrite
             layer.ambient = hsColorRGBA(1.0, 1.0, 1.0, 1.0)
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -166,7 +166,8 @@ class MaterialConverter:
         if slot.use_stencil:
             hsgmat.compFlags |= hsGMaterial.kCompNeedsBlendChannel
             state.blendFlags |= hsGMatState.kBlendAlpha | hsGMatState.kBlendAlphaMult | hsGMatState.kBlendNoTexColor
-            state.clampFlags |= hsGMatState.kClampTexture
+            if slot.texture and slot.texture.type == "BLEND":
+                state.clampFlags |= hsGMatState.kClampTexture
             state.ZFlags |= hsGMatState.kZNoZWrite
             layer.ambient = hsColorRGBA(1.0, 1.0, 1.0, 1.0)
 
--- a/korman/exporter/physics.py
+++ b/korman/exporter/physics.py
@@ -77,14 +77,6 @@ class PhysicsConverter:
             physical.sceneNode = self._mgr.get_scene_node(bl=bo)
 
             getattr(self, "_export_{}".format(bounds))(bo, physical)
-            
-            if self._exporter().has_coordiface(bo):
-                if not physical.mass:
-                    # ...Since the object has a coordinate interface but isn't a kickable, it's likely going to be animated. In such case,
-                    # we must warn Plasma to have the collisions follow the object's coordinates as well.
-                    physical.mass = 1.
-                    simIface.setProperty(plSimulationInterface.kPinned, True)
-                    physical.setProperty(plSimulationInterface.kPinned, True)
         else:
             simIface = so.sim.object
             physical = simIface.physical.object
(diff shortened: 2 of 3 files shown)


korman/exporter/physics.py
```diff
@@ -77,6 +77,14 @@ def generate_physical(self, bo, so, bounds, name=None):
...
 
             getattr(self, "_export_{}".format(bounds))(bo, physical)
+            
+            if self._exporter().has_coordiface(bo):
+                if not physical.mass:
+                    # ...Since the object has a coordinate interface but isn't a kickable, it's likely going to be animated. In such case,
+                    # we must warn Plasma to have the collisions follow the object's coordinates as well.
+                    physical.mass = 1.
+                    simIface.setProperty(plSimulationInterface.kPinned, True)
+                    physical.setProperty(plSimulationInterface.kPinned, True)
```
Hoikas: This is already handled, save the pinned setting, in the aimation modifier's post export. https://github.com/H-uru/korman/blob/master/korman/properties/modifiers/anim.py#L134

korman/exporter/material.py
```diff
@@ -166,7 +166,8 @@ def _export_texture_slot(self, bo, bm, hsgmat, slots, idx):
         if slot.use_stencil:
             hsgmat.compFlags |= hsGMaterial.kCompNeedsBlendChannel
             state.blendFlags |= hsGMatState.kBlendAlpha | hsGMatState.kBlendAlphaMult | hsGMatState.kBlendNoTexColor
-            state.clampFlags |= hsGMatState.kClampTexture
+            if slot.texture and slot.texture.type == "BLEND":
+                state.clampFlags |= hsGMatState.kClampTexture
```
Hoikas: Is it possible to clamp the stencil texture using other buttons? Looking at the code around the diff, I don't see a way that it happens.

korman/exporter/physics.py
Hoikas: Oh, wait, I see what you're saying based on the forum post. Have you tried *only* setting this mass? I'm a little wary of setting random flags just because it works. kPinned does nothing in CWE.

Jrius: Ah, indeed, I missed that part in the animation modifier. It doesn't seem to handle parenting though- that must be the issue, since my test scene has an animated object parenting a physical one. Shoud I move this code to the animation modifier anyway ? To me always setting a mass to objects with a CoordinateInterface sounds safer.
I forgot kPinned only works on Complete Chronicles. It is however required, otherwise the object will start falling due to the nonzero weight (ugh, that sounds so stupid).
How do the ABM and MOUL versions of the engine handle this ? It seems MOUL uses the kPhysAnim flag, but it doesn't prevent the object from falling in CC.

korman/exporter/material.py
Jrius: Do you mean whether the user can choose to clip or not his stencil texture ? Since the stencil is exported as an image afterwards, the "Extension" option in image mapping works just as intended.
My bad about `slot.texture`, I'll correct it.

korman/exporter/physics.py
Hoikas: I think I would handle this in the animation modifier. During the post_export step, it is guaranteed that all plasma objects are created and linked. So, you can check the coordinate interface for the child objects and set the masses on them. I would advise checking all of the detector code and ensuring that the pinned flag is set on all regions. MOUL doesn't need that flag because all detectors are kinematic in MOUL, meaning no forces are applied to the physical.

Jrius:
Latest commit should fix the issues we discussed. Let me know what you think.

korman/properties/modifiers/anim.py
```diff
@@ -143,6 +143,15 @@ def post_export(self, exporter, bo, so):
             # If the mass is zero, then we will fail to animate. Fix that.
             if phys.mass == 0.0:
                 phys.mass = 1.0
+                
+                # On CC and Prime, set kPinned so it doesn't fall through
+                if bpy.context.scene.world.plasma_age.version in ["pvPots", "pvPrime"]:
```
Hoikas: Do this regardless of version--the behavior may change with a physics engine change.

korman/properties/modifiers/anim.py
```diff
@@ -143,6 +143,15 @@ def post_export(self, exporter, bo, so):
...
+                    sim.setProperty(plSimulationInterface.kPinned, True)
+                    phys.setProperty(plSimulationInterface.kPinned, True)
+        
+        # Do the same for children objects
+        for child in so.coord.object.children:
+            self.post_export(exporter, None, child.object)
```
Hoikas: This would work, but I am a little apprehensive about this design. You've assumed here that it is safe to call `post_export` on a modifier more than once. In the future, this `post_export` method might be modified in ways that make it unsafe to call recursively. Or, it might depend on an object having an AGModifier or ATCAnim, which the child objects may not have. I would prefer to see the physics settings stripped into another method and applied from here.

korman/properties/modifiers/anim.py
```diff
@@ -133,6 +133,9 @@ def key_name(self):
 
     def post_export(self, exporter, bo, so):
         # If this object has a physical, we need to tell the simulation iface that it can be animated
+        self.make_physical_movable(so)
+    
+    def make_physical_movable(self, so):
```
Hoikas: One last nitpick :tongue: 

This is a private method, so it should be prefixed by an underscore. Also, everything is sorted by alphabetical order in Korman, so this method should be between `export` and `post_export`

Jrius:
There ! Should be Korman-zen compatible ;)

Hoikas:
Aaaalmost... Just one underscore, not two. :wink:

PR #30, opened 2016-02-15
Lightmap Modifier Refactor

Hoikas:
Refactors the lightmap modifier to limit calls to bake_image as requested on the forum https://forum.guildofwriters.org/viewtopic.php?f=114&t=6663. Also allows for the selection of render layers on the lightmap modifier.

Ping @Deledrius @Jirus


Hoikas:
Updated to fix a bug with vertex colors not being generated.

PR #48, opened 2016-12-29
Implement face flat/smoothing shading

dpogue:
Really I guess this is implementing *non-smoothing*, since we previously treated all faces as if they had smoothing enabled.

It's technically a breaking change, but should make Plasma with runtime lighting match what you'd expect from the Blender 3D viewport.
It also matches how PlasmaMax treats meshes by default with no smoothing groups attached.

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -190,6 +190,7 @@ class MeshConverter:
         for i, tessface in enumerate(mesh.tessfaces):
             data = geodata[tessface.material_index]
             face_verts = []
+            use_smooth = tessface.use_smooth
 
             # Unpack the UV coordinates from each UV Texture layer
             # NOTE: Blender has no third (W) coordinate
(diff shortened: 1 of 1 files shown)


korman/exporter/mesh.py
```diff
@@ -229,7 +229,14 @@ def _export_geometry(self, bo, mesh, materials, geospans):
                     source = mesh.vertices[vertex]
                     geoVertex = plGeometrySpan.TempVertex()
                     geoVertex.position = hsVector3(*source.co)
-                    geoVertex.normal = hsVector3(*source.normal)
+
+                    # If this face has smoothing, use the vertex normal
+                    # Otherwise, use the face normal
+                    if tessface.use_smooth:
```
Hoikas: This is a really tight loop here that is already costly, performance wise. I would like to see assignment of `tessface.use_smooth` to a local value eg `smooth` so that we don't have to look up the `use_smooth` string in `tessface`'s dict every iteration.

Deledrius:
This looks good to me.

PR #49, opened 2016-12-29
Layers should also be clamped with EXTEND

dpogue:
Blender's Texture "Extend" option is what actually matches Plasma's definition of "clamping".

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -488,7 +488,8 @@ class MaterialConverter:
 
             if texture.invert_alpha and has_alpha:
                 state.blendFlags |= hsGMatState.kBlendInvertAlpha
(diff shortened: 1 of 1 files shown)


korman/exporter/material.py
```diff
@@ -488,7 +488,7 @@ def _export_texture_type_image(self, bo, layer, slot):
 
             if texture.invert_alpha and has_alpha:
                 state.blendFlags |= hsGMatState.kBlendInvertAlpha
-        if texture.extension == "CLIP":
+        if texture.extension == "CLIP" or texture.extension == "EXTEND":
```
Hoikas: `if texture.extension in {"CLIP", "EXTEND"}:`

dpogue:
updated

PR #51, opened 2016-12-30
Fix exporting lamp animations

dpogue:
If you had only transform animations on a lamp, it would still attempt to unconditionally export colour/falloff animations.

This early exit was in the LightInfo conversion method, but missing from the OmniLight and SpotLight methods.


Hoikas:
There is a test to make sure animations aren't unconditionally exported, but I am fine with another early exit.

PR #52, opened 2016-12-31
Fix backward lamp distance calculation

dpogue:
When `use_sphere` is enabled, the lamp cutoff is at exactly the distance value. Otherwise, the distance value is the point at which the lamp intensity is half its starting value.

This probably doesn't match Blender 1-to-1, but it's closer to Blender's lighting logic than the existing `use_sphere` calculations.


Hoikas:
Deferring to @Deledrius, who may or may not know more about lighting equations than I do.

PR #54, opened 2017-02-17
fix animation exporter not being able to handle missing axes

MareinK:
I ran into this problem when animating a texture. Normally when setting keyframes, all axes of a transformation are keyed. However if only a subset of all axes is actually manipulated in the animation (and the rest left constant), the keyframes for the other axes can be removed. One reason to do so is to make it clear what aspects are being animated. For example, a water texture that only moves along its X axis does not need keyframes for its Y and Z axes.

In the original code, the assumption was made that all keyframes had X, Y and Z components. Thus an error was raised on line 324, where elements of a dict are accessed that do not exist. By making use of the `pos_default` parameter, these missing values can be filled in by the static values of these properties.

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -320,14 +320,15 @@ class AnimationConverter:
             scale = kwargs.get(scale_path)
 
             # Since only some position curves may be supplied, construct dict with all positions
-            # TODO: probably the same is needed for scale but I don't need it now!
             allpos = dict(enumerate(pos_default))
+            allscale = dict(enumerate(scale_default))
             allpos.update(pos)
+            allscale.update(scale)
 
             matrix = hsMatrix44()
             # Note: scale and pos are dicts, so we can't unpack
             matrix.setTranslate(hsVector3(allpos[0], allpos[1], allpos[2]))
-            matrix.setScale(hsVector3(scale[0], scale[1], scale[2]))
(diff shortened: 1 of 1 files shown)


Hoikas:
This seems like the best way to solve the problem. Would you mind doing the scale as well so we don't have a gotcha lurking in the background?

MareinK:
I tried to apply the same fix for scaling but it has not been a success, see my last commit. For a start a different error popped up that required changes to another part of the code. Having fixed this though, a strange error turns up [on this line](https://github.com/MareinK/korman/blob/c004e5e8f4c6c74c221390813706541f7a11aba7/korman/exporter/animation.py#L331): `TypeError: function takes at most 1 argument (3 given)`, caused by the call to `hsVector3`. I don't  understand this at all since on the line above practically the same call happens successfully. Ideas?

cwalther:
Looking at the [libhsplasma code](https://github.com/H-uru/libhsplasma/blob/master/Python/Math/pyVector3.cpp#L31), it seem like this confusing error message could happen if the three arguments to hsVector3 are not all floating-point numbers. Have you tried printing them out to see what they are?

MareinK:
Thanks for the tip, it turns out my fix for the different error that I mentioned was wrong, and this caused the strange error. Missing axes in both positional and scaling animations are fixed now.

PR #56, opened 2017-06-07
ID Datablock Properties

Hoikas:
:boom: :boom: :boom:

This changeset converts all string object reference properties in Korman to Blender 2.79's new ID Datablock Properties. This will help reduce the amount of boilerplate needed in the exporter around object references as Blender will now take care of sanitizing the data for us. The only downside to this is that the minimum version of Blender is now bumped to 2.79. Additionally, any blendfiles saved with this enabled [will crash older versions of Blender](https://wiki.blender.org/index.php/Dev:Ref/Release_Notes/2.79/Datablocks#Custom_Properties).

Core
- [x] Lamps
- [x] Objects
- [x] Textures
- [x] Worlds

Modifiers
- [x] Animation
- [x] Avatar
- [x] Logic
- [x] Physics
- [x] Region
- [x] Render
- [x] Sound
- [x] Water

Nodes
- [x] Avatar
- [x] Conditions
- [x] Logic
- [x] Message
- [x] Python
- [x] Responder
- [x] Soft Volume

### DIFF
--- a/korman/__init__.py
+++ b/korman/__init__.py
@@ -22,7 +22,7 @@ from . import operators
 bl_info = {
     "name":        "Korman",
     "author":      "Guild of Writers",
-    "blender":     (2, 78, 0),
+    "blender":     (2, 79, 0),
     "location":    "File > Import-Export",
     "description": "Exporter for Cyan Worlds' Plasma Engine",
     "warning":     "beta",
--- a/korman/exporter/etlight.py
+++ b/korman/exporter/etlight.py
@@ -146,12 +146,9 @@ class LightBaker:
 
     def _generate_lightgroup(self, bo, user_lg=None):
         """Makes a new light group for the baking process that excludes all Plasma RT lamps"""
-
-        if user_lg is not None:
-            user_lg = bpy.data.groups.get(user_lg)
         shouldibake = (user_lg is not None and bool(user_lg.objects))
-
         mesh = bo.data
+
         for material in mesh.materials:
             if material is None:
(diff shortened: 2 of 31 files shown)


Hoikas:
The conversion is done. Pull request is ready for review and merge. :smile:

Deledrius:
I have a string of nodes (`PlasmaClickableRegionNode`, `PlasmaClickableNode`, and `PlasmaOneShotMsgNode`) that all seem to lose their object references when I open the .blend with this PR.  References in regular modifiers seem to have been converted correctly.

Deledrius:
It would seem that those values are secretly updating, except one.  Exporting or touching the node tree causes the phantom values to appear.  Only the `PlasmaClickableRegionNode` at the start of the node tree remains unconverted, and even an export fails to fix the missing object reference.

Hoikas:
Blender's Python nodes are very different to the rest of the API, IME. It would not surprise me if in the blender internals, the UI Layout getters do something very naughty, causing the properties to not update as we expect. Fortunately for us, it is much easier to access all of the Plasma nodes as opposed to modifiers. We could iterate through all nodes on blend load and upgrade them without much (if any?) hassle.

Deledrius:
I was hoping you'd say that.  I suppose they were designed more with the API and programmability in mind.

Having that fixed is all this needs to be merged.

Hoikas:
@Deledrius The typos and node tree issues should be resolved now.

Deledrius:
Looks good! 👍

Hoikas:
:boom: :boom: :boom:

PR #58, opened 2017-06-12
Improved export progress logging

Hoikas:
A common complaint that I have heard (and have myself) is that there is little indication of what Korman is doing in the background. Unfortunately, there is no "easy" way to show progress in the Blender UI currently. Korman makes the assumption that nothing will touch Blender's context while it is executing. Attempts to move the exporter into a background thread such that Blender's UI remains active have all resulted in spectacular failures.

In PyPRP, the raw export log was dumped to the console, and many liked to watch it travel by as an indicator of progress. As such, I have implemented a progress monitor that outputs to the console into the export logger. The console now opens during the export (unless disabled in the export operator) to show the export progress. The export operator allows toggling between displaying the simplified progress monitor and the raw export log. If the console was closed before the export, it closes again once the export is finished, unless there is an error. If the console was open, it remains open.

### DIFF
--- a/korlib/texture.cpp
+++ b/korlib/texture.cpp
@@ -266,14 +266,19 @@ static int _generate_detail_map(pyGLTexture* self, uint8_t* buf, size_t bufsz, G
     return 0;
 }
 
-static _LevelData _get_level_data(pyGLTexture* self, GLint level, bool bgra, bool quiet) {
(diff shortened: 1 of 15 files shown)


Hoikas:
I'm going to merge this as it makes little sense to do major work without the new logging facility available.

PR #64, opened 2017-09-11
Add Ladders.

### LINKED ISSUE #44: Ladder modifier
Implement an avatar ladder modifier using MultiStageBehMod

Deledrius:
Adds the plAvLadder modifier, and also a new Plasma submenu to the `shift-A` Add menu.  This menu contains our first premade mesh guide, the Ladder.  

The ladder mesh operator automatically generates guides which show the artist where the climb animation will line up, and the entry regions (if desired) have their plAvLadderMod enabled and pre-configured.

Closes #44.

### DIFF
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -26,7 +26,7 @@ class PlasmaMeshOperator:
 
 class PlasmaAddLadderMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
     bl_idname = "mesh.plasma_ladder_add"
-    bl_label = "Ladder"
+    bl_label = "Add Ladder"
     bl_category = "Plasma"
     bl_description = "Adds a new Plasma Ladder"
     bl_options = {"REGISTER", "UNDO"}
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -1,0 +1,366 @@
+#    This file is part of Korman.
+#
+#    Korman is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    Korman is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
+
+import bpy
+import bmesh
+import math
+import mathutils
+
+class PlasmaMeshOperator:
+    @classmethod
+    def poll(cls, context):
+        return context.scene.render.engine == "PLASMA_GAME"
+
+
+class PlasmaAddLadderMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
+    bl_idname = "mesh.plasma_ladder_add"
+    bl_label = "Ladder"
+    bl_category = "Plasma"
+    bl_description = "Adds a new Plasma Ladder"
+    bl_options = {"REGISTER", "UNDO"}
+
+    # Basic stats
+    ladder_height = bpy.props.FloatProperty(name="Height",
+                                          description="Height of ladder in feet",
+                                          min=6, max=1000, step=200, precision=0, default=6,
+                                          unit="LENGTH", subtype="DISTANCE")
+    ladder_width = bpy.props.FloatProperty(name="Width",
+                                           description="Width of ladder in inches",
+                                           min=30, max=42, step=100, precision=0, default=30)
+    rung_height = bpy.props.FloatProperty(name="Rung height",
+                                          description="Height of rungs in inches",
+                                          min=1, max=6, step=100, precision=0, default=6)
+    # Template generation
+    gen_back_guide = bpy.props.BoolProperty(name="Ladder",
+                                            description="Generates helper object where ladder back should be placed",
+                                            default=True)
+    gen_ground_guides = bpy.props.BoolProperty(name="Ground",
+                                               description="Generates helper objects where ground should be placed",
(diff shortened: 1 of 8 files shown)


korman/operators/op_mesh.py
```diff
@@ -0,0 +1,369 @@
...
+    def draw(self, context):
+        layout = self.layout
+        space = bpy.context.space_data
+        if (not space.local_view):
+            box = layout.box()
+            box.label("Geometry:")
+
+            row = box.row()
+            if (self.ladder_height % 2 != 0):
+                row.alert = True
```
Hoikas: Could be simplified to `row.alert = self.ladder_height % 2 != 0`

korman/operators/op_mesh.py
```diff
@@ -0,0 +1,369 @@
...
+                col.prop(self, "rung_width_type", text="")
+
+            box = layout.box()
+            row = box.row()
+            col = row.column()
+            col.label("Upper Entry:")
+            col.row().prop(self, "has_upper_entry", text="Create")
+            row = col.row()
+            if not self.has_upper_entry:
+                row.enabled = False
```
Hoikas: Could be simplified to `row.enabled = not self.has_upper_entry`

korman/operators/op_mesh.py
```diff
@@ -0,0 +1,369 @@
...
+            row.prop(self, "upper_entry_enabled", text="Enabled")
+            col.separator()
+            col.label("Lower Entry:")
+            col.row().prop(self, "has_lower_entry", text="Create")
+            row = col.row()
+            if not self.has_lower_entry:
+                row.enabled = False
```
Hoikas: Could be simplified to `row.enabled = not self.has_lower_entry`

korman/operators/op_mesh.py
```diff
@@ -0,0 +1,369 @@
...
+        else:
+            row = layout.row()
+            row.label("Warning: Operator does not work in local view mode", icon="ERROR")
+
+    def execute(self, context):
+        if bpy.context.mode == "OBJECT":
+            create_ladder_objects(self, context)
+            return {"FINISHED"}
+        else:
+            self.report({'WARNING'}, "Ladder creation only valid in Object mode")
```
Hoikas: I think it would be a good idea to overload the operator's `poll()` method to do this check as well

korman/properties/modifiers/avatar.py
```diff
@@ -15,13 +15,76 @@
...
+                                    poll=idprops.poll_mesh_objects)
+
+    def export(self, exporter, bo, so):
+        # Create the ladder modifier
+        mod = exporter.mgr.find_create_object(plAvLadderMod, so=so, name=self.key_name)
+        mod.type = plAvLadderMod.kBig
+        mod.loops = self.num_loops
+        mod.enabled = self.is_enabled
+        if self.direction == "UP":
+            mod.goingUp = True
```
Hoikas: Could be simplified to `mod.goingUp = self.direction == "UP"`

korman/operators/op_mesh.py
```diff
@@ -0,0 +1,369 @@
...
+#
+#    You should have received a copy of the GNU General Public License
+#    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
+
+import bpy
+import bmesh
+import math
+import mathutils
+
+class PlasmaMeshCreator:
```
Hoikas: I think the typical naming convention is `PlasmaMeshOperator` IIRC

Deledrius: I think I named this before I figured out where it ultimately ended up.

korman/operators/op_mesh.py
Deledrius: Your logic is reversed, but good idea! :trollface:

korman/operators/op_mesh.py
Deledrius: What would polling for the context mode do in this case?

Hoikas: If the poll fails, the operator should be hidden

Deledrius: It already is, I think?  The Add menu in Edit Mode is a different menu and doesn't show this operator.  The other modes don't even have an Add menu.

I think you're correct, but I'm not sure it's necessary.  Unless I misunderstand the problem.

Hoikas:
Reminder: it looks like there are still some `plasma_object.enabled = True` in here

Deledrius:
I think you're confusing this PR with the journal one.  If the regions aren't Plasma Objects, they won't be exported (and neither will the modifiers on them).

Hoikas:
Crap. Region generation is a pre-export step, right?

Deledrius:
I believe so.  Point is, this is a generator for meshes the artist will interact with.  We're just setting things up, not doing an export here.

korman/ui/modifiers/avatar.py
```diff
@@ -37,3 +37,12 @@ def sittingmod(modifier, layout, context):
     col = split.column()
     col.enabled = modifier.facing_enabled
     col.prop(modifier, "facing_degrees")
+
+def laddermod(modifier, layout, context):
```
Hoikas: Should be above sittingmod

korman/operators/op_mesh.py
```diff
@@ -0,0 +1,366 @@
...
+    def execute(self, context):
+        if bpy.context.mode == "OBJECT":
+            create_ladder_objects(self, context)
+            return {"FINISHED"}
+        else:
+            self.report({"WARNING"}, "Ladder creation only valid in Object mode")
+            return {"CANCELLED"}
+
+
+def create_guide_rungs(self):
```
Hoikas: Should these loose methods be members of the operator's class? They all seem to accept the operator as a self argument?

Deledrius: Good call.

korman/ui/modifiers/avatar.py
Deledrius: 👍

Deledrius:
Updated with your PR, @Hoikas, and I implemented your removed fix for the label as well.

PR #68, opened 2017-10-01
Add support for PythonFileNode attributes arguments.

### LINKED ISSUE #59: DropDownList Attributes
Apparently, `ptAttribDropDownList` support was never added to Python nodes, and it is used by some standard Python files, notably xFogSet.py. Some work will need to be done to store the values in attribute nodes. A bare-bones implementation (sans drop-down) can be found in 9b1a6d7

Deledrius:
This adds the necessary code for the parsing and storage of PFM arguments in the PythonFileNode, for use by the AttributeNodes when appropriate.  

It allows us to do range feedback (Property min and max can't be modified at runtime, so we change the input to an error color) and utilize the strings specified to populate options for DropDownLists (and other list-type nodes to-be-implemented).

It also fixes a missing visitor in the initial parser script which incorrectly skipped negative numbers.

Closes #59.

### DIFF
--- a/korman/nodes/node_python.py
+++ b/korman/nodes/node_python.py
@@ -86,6 +86,52 @@ _attrib_key_types = {
     "ptAttribGrassShader": plFactory.ClassIndex("plGrassShaderMod"),
 }
 
+
+class StringVectorProperty(bpy.types.PropertyGroup):
+    value = StringProperty()
+
+
+class PlasmaAttributeArguments(bpy.types.PropertyGroup):
+    byObject = BoolProperty()
+    default = StringProperty()
+    options = CollectionProperty(type=StringVectorProperty)
+    range_values = FloatVectorProperty(size=2)
+    netForce = BoolProperty()
+    netPropagate = BoolProperty()
+    stateList = CollectionProperty(type=StringVectorProperty)
+    visListId = IntProperty()
+    visListStates = CollectionProperty(type=StringVectorProperty)
+
+    def set_arguments(self, args):
+        for name in args:
+            if name == "byObject":
+                self.byObject = bool(args[name])
+            elif name == "default":
+                self.default = str(args[name])
+            elif name == "options":
+                for option in args[name]:
+                    item = self.options.add()
+                    item.value = str(option)
+            elif name in ("range", "rang"):
+                self.range_values = args[name]
+            elif name == "netForce":
+                self.netForce = bool(args[name])
+            elif name in ("netPropagate", "netProp"):
+                self.netPropagate = bool(args[name])
+            elif name == "stateList":
+                for state in args[name]:
+                    item = self.stateList.add()
+                    item.value = str(state)
+            elif name == "vislistid":
+                self.visListId = int(args[name])
+            elif name == "visliststates":
+                for state in args[name]:
+                    item = self.visListStates.add()
+                    item.value = str(state)
+            else:
+                print("Unknown argument '{}' with value '{}'!".format(name, args[name]))
+
+
 class PlasmaAttribute(bpy.types.PropertyGroup):
     attribute_id = IntProperty()
     attribute_type = StringProperty()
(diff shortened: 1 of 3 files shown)


korman/nodes/node_python.py
```diff
@@ -85,6 +86,50 @@
     "ptAttribGrassShader": plFactory.ClassIndex("plGrassShaderMod"),
 }
 
+class StringVectorProperty(bpy.types.PropertyGroup):
+    value = StringProperty()
+
```
Hoikas: Needs another line of whitespace

korman/nodes/node_python.py
```diff
@@ -85,6 +86,50 @@
...
+class PlasmaAttributeArguments(bpy.types.PropertyGroup):
+    byObject = BoolProperty()
+    default = StringProperty()
+    options = CollectionProperty(type=StringVectorProperty)
+    range = FloatVectorProperty(size=2)
```
Hoikas: Hides builtin function `range`, might not be a good idea

korman/operators/op_nodes.py
```diff
@@ -47,6 +47,47 @@ def invoke(self, context, event):
         context.window_manager.fileselect_add(self)
         return {"RUNNING_MODAL"}
 
+pyAttribArgMap= {
+        "ptAttribute":
+            ["vislistid", "visliststates"],
```
Hoikas: Should these values be sets instead of lists?

korman/operators/op_nodes.py
```diff
@@ -81,5 +122,21 @@ def execute(self, context):
...
+            argmap = {}
+            args = attrib.get("args", None)
+            # Load our default argument mapping
+            if args is not None:
+                if cached.attribute_type in pyAttribArgMap.keys():
+                    argmap.update(dict(zip(pyAttribArgMap[cached.attribute_type], args)))
+                else:
+                    print("Found ptAttribute type '{}' with unknown arguments: {}".format(cached.attribute_type, args))
+            # Add in/set any arguments provided by keyword
+            if not set(pyAttribArgMap[cached.attribute_type]).isdisjoint(attrib.keys()):
```
Hoikas: I think the answer to my above question can be found here...

korman/operators/op_nodes.py
Deledrius: Order is important.  This is the argument order from the function signatures.

korman/operators/op_nodes.py
Deledrius: Checking for a non-empty intersection doesn't care about order, so turning them into sets is fine.

korman/operators/op_nodes.py
Hoikas: Makes sense

korman/nodes/node_python.py
Deledrius: I wondered about that.  I checked that it's not reserved, but I assumed this is why they used `rang` as the keyword.  I assumed it wouldn't be a problem since it's unlikely for the `range` function to apply on this class, but you're right that it's probably best to avoid that assumption and just call it `range_values` or something else irritatingly verbose.

Hoikas: I personally think that the author was awaiting rangarok [sic]...

korman/nodes/node_python.py
```diff
@@ -370,6 +425,31 @@ def update(self):
...
+    bl_label = "Drop Down List Attribute"
+
+    pl_attrib = "ptAttribDropDownList"
+
+    def _list_items(self, context):
+        items = []
+        attrib = self.to_socket
+        if attrib is not None:
+            items = [(option.value, option.value, "") for option in attrib.attribute_arguments.options]
+        return items
```
Hoikas: It might be better to:
```python
if attrib is not None:
     return [(option.value, option.value, "") for option in attrib.attribute_arguments.options]
else:
    return []
```

PR #70, opened 2017-10-02
Set kBlendDetail flag on detail blends

dpogue:
Okay, so this is purely for the sake of technical correctness, because `kBlendDetail` is handled identically to `kBlendAlpha` for rendering purposes


dpogue:
No, it's pretty literally identical: https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/PubUtilLib/plPipeline/DX/plDXPipeline.cpp#L7181-L7184

PR #75, opened 2017-10-08
Fix #63

### LINKED ISSUE #63: Procedural Modifiers and Lightmaps
Currently, static lighting is baked before any object modifiers are applied. This works 99% of the time, however, when procedural modifiers such as the Array modifier are applied, the lighting is only generated for the source data and not for the resulting procedural content.

See also:
https://forum.guildofwriters.org/viewtopic.php?p=71401#p71401

Hoikas:


Deledrius:
I have no visible geometry when I export using this PR.

Hoikas:
Updated to fix fullbright, alldark lightmaps

PR #82, opened 2017-12-05
Fix __exit__ binding to wrong method, ...

zrax:
... and enable the GCC warnings that could catch this kind of thing (unused static function).


Hoikas:
Who needs `__exit__` when you can `__enter__` twice?

Deledrius:
It's just like taking four right turns!

Hoikas:
Wanted to make sure that introducing the correct behavior didn't break anything 🤣

PR #84, opened 2017-12-11
Decal Material support

dpogue:
Decals require 3 things:

* Co-incident geometry that is parented to the main mesh (not currently enforced, but we could warn if you export a decal layer on an object with no parent)
* A material with the CompDecal flag
* Layers with ZIncLayer and NoZWrite

This also reorganizes the layout of the layer properties so provide better grouping:
screenshot from 2017-12-10 22-42-04

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -328,6 +328,9 @@ class MaterialConverter:
                 state.ZFlags |= hsGMatState.kZNoZRead
             if layer_props.skip_depth_write:
                 state.ZFlags |= hsGMatState.kZNoZWrite
+            if layer_props.decal:
+                hsgmat.compFlags |= hsGMaterial.kCompDecal
+                state.ZFlags |= hsGMatState.kZIncLayer | hsGMatState.kZNoZWrite
 
         # Export the specific texture type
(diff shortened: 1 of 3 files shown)


Hoikas:
What does this decal flag give us that the Z Bias flag does not already do?

dpogue:
At a material level, the only difference is that the material will be
affected by the pipeline debug flag to disable decals.

The object parenting is the important part here, which enforces proper
rendering order of decals on top of their parent geometry (including when
that parent is another decal).

This was apparently important enough to warrant its own material type in
PlasmaMax, and a single checkbox is better than needing to remember which
flags to set manually.

Hoikas:
I would prefer to see this as an object modifier as the intended effect is applied at the material level, just like in wavesets and lightmaps. You could add warning messages about parenting to the modifier UI as well.

dpogue:
hmm, yeah, that would make sense. The only question there is if you apply the decal modifier to an object that shares a material with a non-decal object, should the material be duplicated?

Hoikas:
Good question. The way I handled it in wavesets was to ensure that each waveset object had its own material.

Hoikas:
#133 adds a potential solution for this.

PR #86, opened 2017-12-27
Handle shadeless & emit material flags

dpogue:
If the material wants to emit, we'll emit the diffuse colour instead of the world's ambient colour (which pretty much always needs to be black or everything will look wonky)

Also, if we're shadeless, emit white (but do it at runtime via the kShadeWhite flag)

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -733,6 +733,9 @@ class MaterialConverter:
             state.shadeFlags |= hsGMatState.kShadeNoFog # Dead in CWE
             state.shadeFlags |= hsGMatState.kShadeReallyNoFog
 
+        if bm.use_shadeless:
+            state.shadeFlags |= hsGMatState.kShadeWhite
+
         # Colors
         layer.ambient = utils.color(bpy.context.scene.world.ambient_color)
         layer.preshade = utils.color(bm.diffuse_color)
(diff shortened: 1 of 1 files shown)


Deledrius:
Why should a shadeless emission always emit white?

korman/exporter/material.py
```diff
@@ -721,12 +721,24 @@ def _propagate_material_settings(self, bm, layer):
...
+
         # Colors
         layer.ambient = utils.color(bpy.context.scene.world.ambient_color)
         layer.preshade = utils.color(bm.diffuse_color)
         layer.runtime = utils.color(bm.diffuse_color)
         layer.specular = utils.color(bm.specular_color)
 
+        if bm.emit > 0.0:
+            # Use the diffuse colour as the emit, scaled by the emit amount
+            # (maximum 2.0, so we'll also scale that by 0.5)
```
Hoikas: I'm not sure I understand this?

dpogue: Blender Material has an emit slider whose range is apparently [0.0, 2.0]. So a value of 1.0 appears to be 50%, when a multiplication would interpret that as 100%.

dpogue:
> Why should a shadeless emission always emit white?

"emit" in this case means the ambient color. This acts as a starting point before the vertex and diffuse/runtime colours are applied. Typically this is black, but elements like flares can set it to a different colour to ensure they are always fully bright.

Hoikas:
Needs to be rebased before merge

dpogue:
updated

PR #87, opened 2017-12-27
Sort spans for opacity modifiers

dpogue:
Not 100% sure about this, but it was needed in my case to get a lamp flare to render without haloing (in combination with a passIndex to force it into a new BlendSpan).

This is where more rendering modifiers might make sense...


Hoikas:
This makes sense. I had intended at some point to try to figure out these crazy span sorting flags but never got around to it. Thoughts, @Deledrius?

PR #89, opened 2017-12-28
DynaEnvMap fixes

dpogue:
* Export a root node for DynamicEnvMaps, where supported (libHSPlasma handles this, but I think it's only MystV & MOUL)
* Fix a weird vanishing UI field with the DEM visRegion panel

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -518,6 +518,10 @@ class MaterialConverter:
             # will probably want to steal it for diabolical purposes...
             pl_env.position = hsVector3(*viewpt.location)
 
+            # We'll also export it as a Root though, where supported (MOUL)
+            root = self._mgr.find_create_key(plSceneObject, bl=viewpt)
+            pl_env.rootNode = root
+
             if layer is not None:
                 layer.UVWSrc = plLayerInterface.kUVWReflect
                 layer.state.miscFlags |= hsGMatState.kMiscUseReflectionXform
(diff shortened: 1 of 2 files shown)


korman/exporter/material.py
```diff
@@ -518,6 +518,10 @@ def export_dynamic_env(self, bo, layer, texture, pl_class):
             # will probably want to steal it for diabolical purposes...
             pl_env.position = hsVector3(*viewpt.location)
 
+            # We'll also export it as a Root though, where supported (MOUL)
+            root = self._mgr.find_create_key(plSceneObject, bl=viewpt)
+            pl_env.rootNode = root
```
Hoikas: why the duplication?

korman/ui/ui_texture.py
```diff
@@ -38,6 +38,9 @@ def draw(self, context):
         layer_props = context.texture.plasma_layer
         layout = self.layout
 
+        layout.prop(layer_props, "envmap_color")
+        layout.separator()
```
Hoikas: pointless UI ordering change?

dpogue: See the second point, and the full commit message:

> Reorder DEM properties to fix vanishing color prop
>
> Not sure why this was happening, but if you got into a state where no
VisRegion was selected, the Environment Clear Color property would also
disappear.
>
> Reordered to put the Clear Color at the top, so that it works
consistently and also makes it more obvious that it's unrelated to the
visibility regions.

korman/exporter/material.py
dpogue: Mostly because of the comments indicating a desire for the DCM to use a camera viewpoint

Hoikas: Those will require actual cameras to be implemented :wink:

PR #90, opened 2017-12-28
Add support for specular layers

dpogue:
This also allows you to turn off preshade/runtime colouring.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -304,6 +304,18 @@ class MaterialConverter:
             elif slot.blend_type == "MULTIPLY":
                 state.blendFlags |= hsGMatState.kBlendMult
 
+        # Check if this layer uses diffuse/runtime lighting
+        if not slot.use_map_color_diffuse:
+            layer.preshade = hsColorRGBA(0.0, 0.0, 0.0, 1.0)
+            layer.runtime = hsColorRGBA(0.0, 0.0, 0.0, 1.0)
+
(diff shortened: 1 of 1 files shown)


korman/exporter/material.py
```diff
@@ -304,6 +304,18 @@ def export_texture_slot(self, bo, bm, hsgmat, slot, idx, name=None, blend_flags=
...
+        # Check if this layer uses diffuse/runtime lighting
+        if not slot.use_map_color_diffuse:
+            layer.preshade = hsColorRGBA(0.0, 0.0, 0.0, 1.0)
+            layer.runtime = hsColorRGBA(0.0, 0.0, 0.0, 1.0)
+
+        # Check if this layer uses specular lighting
+        if slot.use_map_color_spec:
+            state.shadeFlags |= hsGMatState.kShadeSpecular
+        else:
+            layer.specular = hsColorRGBA(0.0, 0.0, 0.0, 1.0)
```
Hoikas: Magic number?

Deledrius: Do we have premade hsColorRGBA values (e.g. "black")?

dpogue: There are already several spots throughout this file that use "magic colours" (mostly white and black).

In this case, if we're not using specular lighting, the specular colour should be set to black to match PlasmaMax.

Hoikas: Whoops, I misread this as saying specular is always black

Issue #41, opened 2016-06-28
Python packing


Hoikas:
Investigate the ability to create python packages as part of the age export process. This would possibly entail changing the exporter to take a directory and spewing the exported age data there**. See also #34 -- could possibly bundle some nonstandard scripts for PotS compatibility such as the journal template script.

*\* Related: could use Uru directories to parse GlobalAnimations to get animation marker points for OneShot and other logic nodes, preventing common errors.

Deledrius:
This will probably be a necessary addition for distributing Ages easily.  Similar to #34, I'm unsure of the best way to reference the text.  Do we want to use the internal editor for storing Python scripts?  Or externally in a user-specified location?

Creating the packages themselves is fairly trivial.

--
I like the idea of parsing a data dir for the marker points.  If we want to parse locs too, then that solves the question in #34 as well.  We'll probably want a `Rescan` button in the helpers so external changes can be incorporated.  I have a partially-complete branch with a feature that scans the `.age` files for the purposes of exporting new Pages for an existing Age.  Perhaps these should all be consolidated?

Hoikas:
Update... As part of debugging a crash in Doobes' Pub in Path of the Shell, I have determined that PotS will crash when PtGetAgeSDL is called if the AgeSDL blob is missing from the sav file. The creation of this blob has two basic requirements:
1. The presence of `<agename>.sdl` in the SDL directory.
2. The existence of the `<agename>` class in the `<agename>.py[c]` Python file.

(Recall of course that these SDL files use the age name by complete convention of course.  These notes are not intended to be a complete treatise on the functionality of `plPythonSDLModifier`s)

Therefore, this issue needs to be addressed. I see this playing out thusly:
* Update the Python File node to take links to Blender text note blocks.
    * In my testing, Blender will track the local file as well as keep a copy in memory. It notifies the user when it notices the files are out of sync.
    * Text files can be packed. Blender refers to this as making the files "internal".
* Write a new module in pykorlib that uses `ctypes` to load the game's python DLL (`cypython22.dll` or `python27.dll`) for compiling python source code.
    * Unfortunately, MOULa does not ship with a Python 2.3 DLL, AFAICT (but why are you targeting that anyway?)
    * A Python library selector can be added to the games panel for non-Windows systems.
* Write a new stage in the exporter for generating the `<agename>.pak`
    * ABM/PotS will go ahead and xTEA/wdys encrypt the pak and copy it over to the game directory.
[...]

PR #92, opened 2018-06-16
Game tracker

Hoikas:
This begins addressing an ancillary item from #41. Merge as-is or wait for more functionality? Seems to me that the one button press simple export targeting to multiple games is good enough for now.

### DIFF
--- a/korman/operators/op_world.py
+++ b/korman/operators/op_world.py
@@ -36,9 +36,10 @@ class GameAddOperator(AgeOperator, bpy.types.Operator):
         if w:
             # First, verify this is a valid Uru directory...
             path = Path(self.filepath)
+
+            # Blendsucks likes to tack filenames onto our doggone directories...
             if not path.is_dir():
-                self.report({"ERROR"}, "The selection is not a valid directory.")
-                return {"CANCELLED"}
+                path = path.parent
             if not ((path / "UruExplorer.exe").is_file() or (path / "plClient.exe").is_file()):
                 self.report({"ERROR"}, "The selected directory is not a copy of URU.")
                 return {"CANCELLED"}
@@ -53,7 +54,7 @@ class GameAddOperator(AgeOperator, bpy.types.Operator):
                 game = games.games[self.game_index]
 
             # Setup game...
-            game.path = self.filepath
+            game.path = str(path)
             if (path / "cypython22.dll").is_file():
                 game.version = "pvPots"
             else:
--- a/korman/operators/op_world.py
+++ b/korman/operators/op_world.py
@@ -14,12 +14,76 @@
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
 import bpy
-
+from bpy.props import *
+from pathlib import Path
 
 class AgeOperator:
     @classmethod
     def poll(cls, context):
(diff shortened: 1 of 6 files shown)


dpogue:
Does this require exporting into the game folder? I always export to a working folder, and then make a bunch of manual file modifications before copying the files into the game folder.

Hoikas:
The new functionality requires exporting into a game folder; however, the ability to export loose age files from the export menu has not been removed.

Deledrius:
Excellent.  Even though I usually export directly to my game folder, I would hate to lose the bare-export functionality.

korman/properties/prop_world.py
```diff
@@ -158,6 +191,10 @@ def export(self, exporter):
     use_texture_page = BoolProperty(name="Use Textures Page",
                                     description="Exports all textures to a dedicated Textures page",
                                     default=True)
+    age_name = StringProperty(name="Age Name",
+                              description="Name of the Age to be used for data files",
+                              default="Korman Age",
```
Deledrius: Might be a good idea to leave this default blank, and marked invalid (red).  I didn't notice this field had been added and it even took me a moment to find it in the UI since it was already filled in with a name.  I've been exporting my Age with the wrong name!

korman/operators/op_world.py
```diff
@@ -14,14 +14,78 @@
...
+    filepath = StringProperty(subtype="DIR_PATH")
+    directory = BoolProperty(default=True, options={"HIDDEN"})
+    game_index = IntProperty(default=-1, options={"HIDDEN"})
+
+    def execute(self, context):
+        w = context.world
+        if w:
+            # First, verify this is a valid Uru directory...
+            path = Path(self.filepath)
+            if not path.is_dir():
```
Deledrius: Unfortunately Blender likes to auto-fill the filename field in the selector dialog, so we should strip that portion to avoid confounding error messages when the user selected a directory and Blender is too "helpful".

korman/ui/ui_world.py
```diff
@@ -26,6 +29,47 @@ def poll(cls, context):
...
+            box.prop(active_game, "version")
+            box.separator()
+
+            row = box.row(align=True)
+            op = row.operator("world.plasma_game_add", icon="FILE_FOLDER", text="Change Path")
+            op.filepath = active_game.path
+            op.game_index = active_game_index
+            row.operator_context = "EXEC_DEFAULT"
+            op = row.operator("export.plasma_age", icon="EXPORT")
+            op.filepath = str((Path(active_game.path) / "dat" / age.age_name).with_suffix(".age"))
```
Deledrius: When age_name is blank, we end up with a nice export of dat.age in the game's root folder. :trollface: We should probably refuse to export and give a pop-up error instead.

korman/ui/ui_world.py
```diff
@@ -77,10 +121,26 @@ def draw(self, context):
         col = split.column()
         col.label("Age Settings:")
         col.prop(age, "seq_prefix", text="ID")
+        col.prop(age, "age_name", text="")
```
Deledrius: We should set the `alert` property on this one if the value is currently empty, so that it calls attention to the invalid name.

PR #94, opened 2018-06-17
Fix ladder detector regions for PotS exports.

### LINKED ISSUE #93: Ladder Modifier regions export incorrectly on PotS.
As [reported by J'Kla on the forum](https://forum.guildofwriters.org/viewtopic.php?f=10&t=6875), the new Ladder Modifiers do not export correctly for use in The Path of the Shell.  Both the top and bottom regions appear to be at the Age's origin instead of their correct positions.  I've confirmed the bug, but cannot determine the cause.

Deledrius:
Apparently pre-MOUL ladders can't have massless regions as detectors anywhere but at the origin, so this adds the required values to make sure the export puts each one in the correct location in PotS.

Fixes #93.

### DIFF
--- a/korman/properties/modifiers/avatar.py
+++ b/korman/properties/modifiers/avatar.py
@@ -77,6 +77,10 @@ class PlasmaLadderModifier(PlasmaModifierProperties):
         simIface, physical = exporter.physics.generate_physical(bo, so, bounds, det_name)
         physical.memberGroup = plSimDefs.kGroupDetector
         physical.reportGroup |= 1 << plSimDefs.kGroupAvatar
+        physical.setProperty(plSimulationInterface.kPinned, True)
+        simIface.setProperty(plSimulationInterface.kPinned, True)
+        if physical.mass == 0.0:
+            physical.mass = 1.0
 
     @property
     def requires_actor(self):


korman/properties/modifiers/avatar.py
```diff
@@ -77,6 +77,8 @@ def export(self, exporter, bo, so):
         simIface, physical = exporter.physics.generate_physical(bo, so, bounds, det_name)
         physical.memberGroup = plSimDefs.kGroupDetector
         physical.reportGroup |= 1 << plSimDefs.kGroupAvatar
+        physical.setProperty(plSimulationInterface.kPinned, True)
```
Hoikas: This needs to be set on the `plSimulationInterface` as well. It's functionally a no-op, but I prefer for the bit flags to be identical on the two objects.

korman/properties/modifiers/avatar.py
```diff
@@ -77,6 +77,8 @@ def export(self, exporter, bo, so):
...
+        physical.mass = 1.0
```
Hoikas: The convention for doing this is:
```python
    if physical.mass == 0.0:
        physical.mass = 1.0
```

Deledrius: Can do.

korman/properties/modifiers/avatar.py
Deledrius: Done.

PR #95, opened 2018-06-18
Deprecated Responder Command Nodes

### LINKED ISSUE #36: Improve Responder node workflow
Work toward deprecating the Responder State node and move toward implicit responder states built from command linkage. This will ideally make the logic node tree easier to follow. Will also need to build in a framework for detecting deprecated node types on load and upgrading them on-the-fly. But that shouldn't be too hard.

Hoikas:
Fixes #36


Hoikas:
This needs some adjustments as it is possible to specify impossible Responders in the node graph.

Issue #96, opened 2018-07-06
ImageLibMods


Hoikas:
The journal modifier should be updated to allow the user to list images they would like to be able to use in the journal's PBML. This should automatically export a plImageLibMod so those images will be properly loaded by the engine.

Deledrius:
I agree.  This sounds like a good idea.

PR #103, opened 2018-08-05
Fix #76

### LINKED ISSUE #76: Export lightmaps as PNG
CWE now supports PNG mipmaps. Therefore, it would be a filesize win for us to compress lightmaps as PNGs for export to MOUL.

Hoikas:
On the surface, this allows lightmaps to be exported as PNGs. However, it also includes a light refactor of image exporting to facilitate requesting export formats and the targeting of plImageLibMod. This should better enable #96.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -45,10 +45,11 @@ class _Texture:
         else:
             self.layer = kwargs.get("layer")
             self.calc_alpha = False
-            self.mipmap = False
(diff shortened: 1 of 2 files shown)


Deledrius:
Looks good.  Works on lightmaps, and is exactly what I need for #96.

Hoikas:
This is currently blocking on BGR fixes to libHSPlasma.

PR #107, opened 2018-08-18
Responder State IDs

### LINKED ISSUE #100: Responder State IDs
Presently, Responder State nodes are linked to a single output node on the Responder. Additionally, they can be linked to one another and theoretically be included in the Responder. While this is sufficient for general use cases, for Responder states to be used in Python via `ptAttribResponder`, the Responder State's index must be known by the Python script.

I envision this being fixed by removing the States output node and replacing it with a single node socket that only accepts one state. The socket will report the Responder State's index for use in scripts. Additionally, states MUST be connected to the Responder's output in order to be exported. States can still trigger other states, however, this will be done via a different node socket.

Hoikas:
This changeset enables Responder nodes to track Responder State nodes individually and report their ID to the user for easier integration into Python scripts. This seems like a good companion for #95 in Korman 0.07

Fixes #100

### DIFF
--- a/korman/nodes/node_core.py
+++ b/korman/nodes/node_core.py
@@ -113,11 +113,31 @@ class PlasmaNodeBase:
                         continue
                     yield node
 
-    def find_output_socket(self, key):
+    def find_output_socket(self, key, spawn_empty=False):
+        # In the case that this socket will be used to make new output linkage,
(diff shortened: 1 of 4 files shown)


Deledrius:
I think I have a scripted door ready to test this on. ;)

Hoikas:
Since you volunteered :wink:

PR #109, opened 2018-08-19
Fix #106

### LINKED ISSUE #106: Limit export to objects in scene
Korman currently exports all objects in a blend file marked as Plasma Object. With the introduction of ID Properties, objects removed from the scene might not be removed from the blend due to references from other objects. As part of the export, we need to ensure only objects in valid scenes are exported. We should probably also investigate warning if we try to export a reference to an object that is either not a plasma object or not in a valid scene.

See also https://forum.guildofwriters.org/viewtopic.php?f=10&t=6946

Hoikas:
Given that the Plasma World (read: age) is defined per scene, it seems appropriate to limit the export to the current scene. Other scenes could potentially have other age definitions...

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -112,12 +112,13 @@ class Exporter:
         oven.bake_static_lighting(self._objects)
 
     def _collect_objects(self):
+        scene = bpy.context.scene
         self.report.progress_advance()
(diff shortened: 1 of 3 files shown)


Deledrius:
Scenes being almost entirely independent things in Blender would suggest this makes the most sense as to how Korman should handle them.  I don't use them at all in my current Age building workflow.

PR #112, opened 2018-08-21
Allow version specific pages

Hoikas:
This allows the artist to set specific export versions for PRPs, enabling game version specific age features.


Deledrius:
Seems to work on my trivial testing Age.  Nice! 👍

PR #122, opened 2018-09-09
Add named lightmap passes

### LINKED ISSUE #118: Named Lightmap Passes
When creating a larger age with many lightmaps, it can become difficult to manage lightmap modifiers. This is because lightmap modifiers currently expect for the artist to specify which layers are active on each lightmap modifier. A better solution would be to maintain a collection of named lightmap passes such that changes to each pass would propagate to the other lightmaps in said pass. This will require functionality for versioning and upgrading modifiers.

Hoikas:
This fixes #118 by adding named lightmap passes. It is no longer possible to specify render layers on a per-object basis. Instead, a pass must be set up. Old modifiers are silently upgraded and consolidated into passes when the blend file is loaded. The old default behavior is equivalent to specifying no pass on the new modifier.

Additionally, I have generalized the lightmap modifier by allowing the artist to change to vertex shading mode and specify the layers to bake with. The "bake static lighting" checkbox has been changed to a dropdown in which light baking can be turned off completely, all lightmaps can be forced to vertex colors, all arist-specified vertex colors can be forced to lightmaps, or the blissful default. I did not add the option to force auto-vertex colors to lightmaps, which I suppose could be interesting in some situations...

### DIFF
--- a/korman/exporter/etlight.py
+++ b/korman/exporter/etlight.py
@@ -221,6 +221,7 @@ class LightBaker(_MeshManager):
         # that needs to be baked and sort it out by configuration.
         bake = { ("vcol",): [] }
         bake_vcol = bake[("vcol",)]
+        bake_passes = bpy.context.scene.plasma_scene.bake_passes
 
         for i in objs:
             if i.type != "MESH":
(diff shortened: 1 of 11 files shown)


Deledrius:
This fixes some of the weird vertex lighting issues I was having, I assume by allowing me to force it to rebake on certain objects.

korman/exporter/etlight.py
```diff
@@ -231,15 +233,24 @@ def _harvest_bakable_objects(self, objs):
             mods = i.plasma_modifiers
             lightmap_mod = mods.lightmap
             if lightmap_mod.enabled:
+                if lightmap_mod.bake_pass_name:
+                    bake_pass = bake_passes.get(lightmap_mod.bake_pass_name, None)
+                    if bake_pass is None:
+                        raise ExportError("Bake Lighting '{}': Could not find pass '{}'".format(i.name, lightmap_mod.bake_pass_name))
+                    lm_layers = tuple(bake_pass.render_layers)
```
Deledrius: It would be nice to save the current layer visibility and restore them at the end.  Otherwise, we potentially enable a lot of layers (by default) that weren't enabled previously.

PR #124, opened 2018-10-24
Add MultiStageBehMod Node.

### LINKED ISSUE #45: MultiStageBehMod logic nodes
These are required for complex behaviors

Deledrius:
Add MultiStage Behaviour, AnimationStage, and SeekTarget nodes.

Closes #45.

### DIFF
--- a/korman/nodes/node_avatar.py
+++ b/korman/nodes/node_avatar.py
@@ -20,12 +20,13 @@ from PyHSPlasma import *
 
 from .node_core import PlasmaNodeBase, PlasmaNodeSocketBase
 from ..properties.modifiers.avatar import sitting_approach_flags
+from ..exporter.explosions import ExportError
 
 class PlasmaSittingBehaviorNode(PlasmaNodeBase, bpy.types.Node):
     bl_category = "AVATAR"
     bl_idname = "PlasmaSittingBehaviorNode"
     bl_label = "Sitting Behavior"
-    bl_default_width = 100
+    bl_width_default = 120
 
     approach = EnumProperty(name="Approach",
                             description="Directions an avatar can approach the seat from",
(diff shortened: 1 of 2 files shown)


korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+            seek_target = seek_socket.links[0].from_node.target
+        elif seek_socket.auto_target:
+            smartSeek = True
+            seek_target = bo
+        else:
+            smartSeek = False
+            seek_target = bo
+
+        # We need to make sure the object we're attached to is being exported as a Plasma Object, otherwise this is all nonsense
+        seek_target.plasma_object.enabled = True
```
Hoikas: The exporter will proceed just fine without it being marked Plasma Object. The other modifiers on the object just won't be exported.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+
+        if smartSeek:
+            spm = exporter.mgr.find_create_object(plSeekPointMod, name=seek_target.name, bl=seek_target)
+            spm.name = seek_target.name
+
+        msbmod = self._find_create_object(plMultistageBehMod, exporter, bl=seek_target)
```
Hoikas: Since this is a single modifier, we need to be really careful about how we export it. This snippet will prevent the MSB from being exported more than once, duplicating the stages:
```python
    @property
    def export_once(self):
        return self.find_input_socket("seek_target").is_linked
```

Additionally, the pattern used by most nodes to get the "publlic" object is to call its own `get_key` method and grab the object from there. This is helpful for when code inevitable has to be tweaked :wink:

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+                    animstage.regressTo = None
+                else:
+                    # Don't Auto-Regress, just stop!
+                    animstage.regressTo = -1
+
+            msbmod.addStage(animstage)
+
+    @property
+    def requires_actor(self):
+        return True
```
Hoikas: This will cause a CI to be exported for the MSB even when the seek point not the object the tree is attached to. Additionally, a CI is not requested for an explicit seek point. This will generally work because I expect empties will be used, but sometimes you get people like me who use a mesh for these things...

```python
    @property
    def requires_actor(self):
        return not self.find_input_socket("seek_target").is_linked

    @property
    def harvest_actors(self):
        seek_socket = self.find_input_socket("seek_target")
        if seek_socket.is_linked:
            yield seek_socket.links[0].from_node.target.name

File: korman/nodes/node_avatar.py [hunk 4]
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+            "text": "Host Script",
+            "type": "PlasmaBehaviorSocket",
+            "valid_link_nodes": {"PlasmaPythonFileNode"},
+            "spawn_empty": True,
+        })
+    ])
+
+    def draw_buttons(self, context, layout):
+        col = layout.column()
+        col.label("Options:")
```
Hoikas: The other nodes that have options like these don't have this label :wink:

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+class PlasmaSeekTargetSocket(PlasmaNodeSocketBase, bpy.types.NodeSocket):
+    bl_color = (0.180, 0.350, 0.180, 1.0)
+
+
+class PlasmaSeekTargetSocketIn(PlasmaSeekTargetSocket):
+    auto_target = BoolProperty()
+
+    def draw(self, context, layout, node, text):
+        if not self.is_linked:
+            layout.prop(self, "auto_target", text="Auto Smart Seek")
```
Hoikas: It would be better, I think, to add the text and a description directly on the property.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+        if self.loop_option == "kLoop":
+            row = col.row()
+            row.prop(self, "num_loops")
+        col.separator()
+
+    @property
+    def stage_id(self):
+        idx = None
+        if self.find_output_socket("stage").is_linked:
+            parent_MSB = self.find_output_socket("stage").links[0].to_node
```
Hoikas: It would be a small performance win to only evaluate `self.find_output_socket("stage")` once.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+        ("stage", {
+            "text": "Stage",
+            "type": "PlasmaAnimStageSettingsSocket",
+            "valid_link_nodes": "PlasmaAnimStageNode",
+            "valid_link_sockets": "PlasmaAnimStageSettingsSocket",
+        }),
+    ])
+
+    def draw_buttons(self, context, layout):
+        col = layout.column()
```
Hoikas: it's already laid out in a column?

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+        ("stage_reference", {
+            "text": "Stage Progression",
+            "type": "PlasmaAnimStageOrderSocketOut",
+            "valid_link_nodes": "PlasmaAnimStageSettingsNode",
+            "valid_link_sockets": {"PlasmaAnimStageAdvanceSocketIn", "PlasmaAnimStageRegressSocketIn"} ,
+        }),
+    ])
+
+    def draw_buttons(self, context, layout):
+        col = layout.column()
```
Hoikas: already a column?

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+        ("hosts", {
+            "text": "Host Script",
+            "type": "PlasmaBehaviorSocket",
+            "valid_link_nodes": {"PlasmaPythonFileNode"},
+            "spawn_empty": True,
+        })
+    ])
+
+    def draw_buttons(self, context, layout):
+        col = layout.column()
```
Hoikas: ?

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+    auto_advance = BoolProperty(name="Advance to Next Stage",
+                                description="Automatically advance to the next stage when the animation completes instead of halting",
+                                default=True)
+
+    def draw(self, context, layout, node, text):
+        if not self.is_linked:
+            col = layout.column()
+            col.prop(self, "auto_advance")
+        else:
+            if self.links[0].from_node.stage_id is not None:
```
Hoikas: Would be nice to have a local for this to decrease the verbosity.

korman/nodes/node_avatar.py
Deledrius: Ah, I did not know this!  There was more logic involved here, before we made the decision to assume we're attached to the object we're going to export on, so it's a bit redundant besides, I suppose.

korman/nodes/node_avatar.py
Deledrius: I saw that in your recent PR, but wasn't sure if it was needed here.

korman/nodes/node_avatar.py
Deledrius: Good catch.  I think this was a leftover from the early mockup stage.  💥

korman/nodes/node_avatar.py
Deledrius: Why check if the seek_target is linked for this?  Shouldn't it check the `host` socket connection to the PFM?

korman/nodes/node_avatar.py
Deledrius: Excellent, thanks!  I was unsure about this part.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +71,388 @@ def export(self, exporter, bo, so):
...
+            seek_target = bo
+        else:
+            smartSeek = False
+            seek_target = bo
+
+        # We need to make sure the object we're attached to is being exported as a Plasma Object, otherwise this is all nonsense
+        seek_target.plasma_object.enabled = True
+
+        if smartSeek:
+            spm = exporter.mgr.find_create_object(plSeekPointMod, name=seek_target.name, bl=seek_target)
```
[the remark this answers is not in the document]
Hoikas: Ok, after needing to think about this, I think I understand now. This modifier needs to be attached to the same SO as the MSB to specify the seek point. AFAICT though if you leave it out, the seek point is considered to be the object the MSB itself is attached to. So, at the very least, this needs to be using `self._find_create_object(plSeekPointMod, exporter, **kwargs)`.

korman/nodes/node_avatar.py
Hoikas: The seek target socket appears to be triggering whether or not we export a single MSB (seek point set) or multiple MSBs (seek point is the attached object).

korman/nodes/node_avatar.py
Deledrius: Oh, bizarre.  I must have just thrown this together (back during the previous run at this code) for testing.  Good catch, I'll put that back where it belongs.

korman/nodes/node_avatar.py
Deledrius: `TypeError: 'generator' object is not callable`

Hoikas: hwat?

Hoikas: I see now. I updated the snippet.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,395 @@ def export(self, exporter, bo, so):
...
+        if not self.is_linked:
+            layout.prop(self, "auto_regress")
+        else:
+            if self.links[0].from_node.stage_id is not None:
+                layout.label("{} {}".format(text, self.links[0].from_node.stage_id))
+            else:
+                layout.label(text)
+
+
+class PlasmaAnimStageOrderSocketOut(PlasmaAnimStageOrderSocket):
```
Hoikas: Best to have any classes used by blender inherit the blender base directly. Meaning this should inherit from `bpy.types.NodeSocket` explicitly, instead of implicitly via `PlasmaAnimStageOrderSocket`. Interesting issues occur, mainly in properties, if you expect for inheritance to work properly with Blender classes.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,395 @@ def export(self, exporter, bo, so):
...
+                layout.label("Stage (ID: {})".format(idx))
+        else:
+            layout.label(text)
+
+
+class PlasmaSeekTargetSocket(PlasmaNodeSocketBase, bpy.types.NodeSocket):
+    bl_color = (0.180, 0.350, 0.180, 1.0)
+
+
+class PlasmaSeekTargetSocketIn(PlasmaSeekTargetSocket):
```
Hoikas: See comment about socket inheritance.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,395 @@ def export(self, exporter, bo, so):
...
+            layout.prop(self, "auto_target")
+        else:
+            target = self.links[0].from_node.target
+            if target:
+                layout.label("Smart Seek Target: {}".format(target.name))
+            else:
+                layout.label("Smart Seek Target")
+
+
+class PlasmaSeekTargetSocketOut(PlasmaSeekTargetSocket):
```
Hoikas: See comment about socket inheritance.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,395 @@ def export(self, exporter, bo, so):
...
+class PlasmaSeekTargetSocket(PlasmaNodeSocketBase, bpy.types.NodeSocket):
+    bl_color = (0.180, 0.350, 0.180, 1.0)
+
+
+class PlasmaSeekTargetSocketIn(PlasmaSeekTargetSocket):
+    auto_target = BoolProperty(name="Auto Smart Seek",
+                               description="Smart Seek causes the avatar to seek to the provided position before starting the behavior ('auto' will use the current object as the seek point)",
+                               default=False)
+
+    def draw(self, context, layout, node, text):
```
Hoikas: This will override the ability for `PlasmaNodeSocketBase` to draw the + operator button. The method should be named `draw_content` instead.

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,395 @@ def export(self, exporter, bo, so):
...
+    def harvest_actors(self):
+        seek_socket = self.find_input_socket("seek_target")
+        if seek_socket.is_linked:
+            yield seek_socket.links[0].from_node.target.name
+
+
+class PlasmaAnimStageRefSocket(PlasmaAnimStageSocket, bpy.types.NodeSocket):
+    bl_color = (0.188, 0.186, 0.349, 1.0)
+
+    def draw(self, context, layout, node, text):
```
Hoikas: This will override the ability for `PlasmaNodeSocketBase` to draw the + operator button. The method should be named `draw_content` instead.

korman/nodes/node_avatar.py
Deledrius: Oh, that's why it didn't work on this one!

korman/nodes/node_avatar.py
Deledrius: Oh, that's why it didn't work on this one!

korman/nodes/node_avatar.py
```diff
@@ -71,3 +72,393 @@ def export(self, exporter, bo, so):
...
+            seek_target = bo
+        else:
+            smartSeek = False
+            seek_target = bo
+
+        if smartSeek:
+            spm = self._find_create_object(plSeekPointMod, exporter, bl=seek_target)
+            spm.name = seek_target.name
+
+        msbmod = self._find_create_object(plMultistageBehMod, exporter, bl=seek_target)
```
Hoikas: I somehow missed that this still doesn't use `get_key` like it should...

Deledrius: `_find_create_object` doesn't take a key, does it?

korman/nodes/node_avatar.py
Deledrius: That's unfortunate.

korman/nodes/node_avatar.py
Hoikas: This is still using naughty socket inheritance.

korman/nodes/node_avatar.py
Hoikas: This is still using naughty socket inheritance.

korman/nodes/node_avatar.py
Deledrius: Ugh, missed these, got the Stage Ordering sockets. >_<

Hoikas:
Thanks for getting this in!

Deledrius:
Sorry it took so ridiculously long!

PR #128, opened 2019-01-02
Python Packing

### LINKED ISSUE #127: Move Plasma Games data storage
Given that the location of Plasma game executables is a per-computer type setting, we should move the Plasam Games settings from `bpy.types.World` to a Korman [bpy.types.AddonPreferences](https://docs.blender.org/api/blender_python_api_current/bpy.types.AddonPreferences.html).

Hoikas:
Welcome to the changeset where everything is changed and the rules don't matter! :boom:

This does a lot of things to add the ability to compyle and pack age Python. First, the machinery around exporting files is completely redone such that it is no longer legal to open a file for writing anywhere in the exporter aside from requesting it from the newfangled output file manager. The output file manager decides what kind of stream you get, including if it is even backed by a real file. To test this, it is now possible to generate a zipfile of your age. To generate a zip of only the age/prp files, you can use the File > Export menu operator and export `<agename>.zip`. To get a complete age image, a new button is provided in the World panel.

Python files are compyled into `<agename>.pak` in a few cases. First, the python file must be backed by a `bpy.types.Text` datablock. These are great because they store the file in Blender's memory so exchanging blendfiles becomes a bit easier. Second, the python file must be marked as "Package". This can be done either in the text editor screen or by clicking the "Pack" button on a Python File node. If a Python File node's backing Text datablock is marked Package, it will be added to the Python pak with the glue appended. If a non-Python File Node python file is found in the blendfile and marked "Package", it will still be added to the package but without the glue appended. BUT REMEMBER! The rules don't matter, so in MOUL the python code is just copied to the export directory and not packed.

To facilitate exchanging blendfiles among artists and working on mutliple ages, game install specification has moved into the Korman addon preferences. Python compylation is done by using Python installs specified here as well. In the most common use case (Windows with Korman installer), Python 2.2 will be installed by our installer and found by querying the registry.

There is more to this changeset, but I am le tired.

Closes #127 and #41


korman/plasma_magic.py
```diff
@@ -0,0 +1,225 @@
...
+    VERSION 0
+}}
+
+"""
+
+# Copypasta (with small fixes for str.format) of glue.py from CWE's moul-scripts
+plasma_python_glue = """
+# glue code in python for the glue code in C++
+# This assumes that this will be loaded into the module that we are trying to do
+# with an execfile('.\\python\\system\\glue.py') at the end of the module (after everything is defined)
```
Deledrius: I see that this is an error in the original file, but the path is incorrect.  It's currently found in [Python/plasma/glue.py](https://github.com/H-uru/moul-scripts/blob/master/Python/plasma/glue.py).

korman/exporter/convert.py
```diff
@@ -107,6 +107,11 @@ def run(self):
                 self.report.progress_end()
                 self.report.save()
 
+                # Step 5.2: If any nonfatal errors were occured during the export, we will
```
Deledrius: The phrase "were occured" is strange (a combination of "were encountered" and "occurred" perhaps?).  Just "occurred" should be sufficient.

korman/exporter/outfile.py
```diff
@@ -188,13 +222,46 @@ def _generate_files(self, func=None):
...
+            for i in self._generate_files(func):
+                if i.needs_glue:
+                    py_code = "{}\n\n{}\n".format(i.file_data, plasma_python_glue)
+                else:
+                    py_code = i.file_data
+                result, pyc = korlib.compyle(i.filename, py_code, py_version, report, indent=1)
+                if result:
+                    pyc_objects.append((i.filename, pyc))
+        except korlib.PythonNotAvailableError as error:
+            report.warn("Python {} is not available. Your age scripts were not packaged.", error, indent=1)
```
Deledrius: Style: Age should be capitalized.

korman/exporter/outfile.py
```diff
@@ -206,9 +273,31 @@ def save(self):
         else:
             self._write_deps()
 
+    @property
+    def super_secure_encryption(self):
+        version = self._version
+        if version == pvEoa:
+            return plEncryptedStream.kEncAes
+        elif version == pvMoul:
+            # trollface.jpg
```
Deledrius: It's probably a discussion for another time, but this is likely the best option.  The encryption key is server-specific at the MO:UL subversions, and should be handled by the server operator not the Age builder.

korman/korlib/python.py
```diff
@@ -0,0 +1,200 @@
...
+            if _verify_python(py_version, py_executable):
+                return py_executable
+
+        # Second, try looking Python up in the registry.
+        try:
+            import winreg
+        except ImportError:
+            pass
+        else:
+            py_executable = _find_python_reg(winreg.HKEY_LOCAL_MACHINE, py_version)
```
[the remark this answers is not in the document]
Hoikas: I was probably thinking about how the Python 2.2 installer uses HKLM by default

korman/plasma_magic.py
Hoikas: I should probably just remove the comments entirely... The code is compiled and the comments trashed :wink:

korman/exporter/outfile.py
Hoikas: Indeed. Currently, I just outright disallowed building a python pak for MOUL/CWE because that engine only loads python.pak. Maybe if we ever get a more distributed system with true multiple shard support....

Hoikas:
Fixed issues from review

korman/korlib/__init__.py
```diff
@@ -101,3 +103,35 @@ def inspect_wavefile(stream, header):
         header.read(stream)
 
         return chunks[b"data"]["size"]
+
+    def is_legal_python2_identifier(identifier):
```
Deledrius: I think using the [built-in function for this](https://docs.python.org/3/library/stdtypes.html#str.isidentifier) might be sufficient.  According to [the docs](https://docs.python.org/3/reference/lexical_analysis.html#identifiers), the only additional restriction in Python 2 is that the string must also be [within the ASCII range](https://docs.python.org/3/library/stdtypes.html#str.isascii):

> Within the ASCII range (U+0001..U+007F), the valid characters for identifiers are the same as in Python 2.x: the uppercase and lowercase letters A through Z, the underscore _ and, except for the first character, the digits 0 through 9.
> 
> Python 3.0 introduces additional characters from outside the ASCII range (see PEP 3131). For these characters, the classification uses the version of the Unicode Character Database as included in the unicodedata module.

In which case, I believe this entire function can be reduced to:

`return str.isidentifier(identifier) and str.isascii(identifier)`

Hoikas: Unfortunately, we are presently constrained to Python 3.5. `identifier.isascii()` was added in Python 3.7. But, also, in the documentation right next door is a note about keywords. I did forget to ensure someone isn't trying to name their age `class`, `def`, `if`, `else`, `yield`, etc...

Hoikas:
Updated to include a less verbose identifier checker and protection for naming ages after Python keywords.

korman/korlib/__init__.py
Deledrius: Ah, strange that `isascii` is so new.  Still, that leaves a lot of simplification!

cwalther: Isn’t that just a more convoluted way of saying
```python
if any(ord(i) > 0x7F for i in identifier):
    return False
```
?

Hoikas:
:bomb:

3... 2... 1...

:boom: :boom: :boom: :boom: :boom: :boom:

PR #130, opened 2019-01-14
Fix #129

### LINKED ISSUE #129: Journal Modifier Data Packing
Continuation from #34.

With the incorporation of #128, the Journal modifier needs to be revisited to pack journal data into the export properly. Careful consideration is needed for how localization should be handled in MOUL/CWE. Do we prefer generation of locs from text blocks or would we rather see something like a localization editor?

Hoikas:
One usability question. Currently, we present the languages in their native-type name. Perhaps we should use their English name instead?

### DIFF
--- a/korman/exporter/__init__.py
+++ b/korman/exporter/__init__.py
@@ -18,5 +18,6 @@ from PyHSPlasma import *
 
 from .convert import *
 from .explosions import *
+from .locman import *
 from .python import *
 from . import utils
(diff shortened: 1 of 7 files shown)


Deledrius:
I think using the native name is probably best.  I know it makes it easier for me to find English in a list when it's in English.

Hoikas:
English is the default selection BTW

Hoikas:
Updated to include the ability to export localization data separately from the age export process.

PR #133, opened 2019-01-27
Fix #132

### LINKED ISSUE #132: Shared lightmap on multi-user materials
Heya !
On the 0.07 version of Korman, it seems lightmaps are baked directly to the material(s) the object is using. This will cause issues if another object uses the same material (resulting in the second object using the lightmap of the first object). Korman should leave the original material untouched, and instead bake the lightmap to a copy of this material. For now, I can work around this issue by making a single-user copy of the material in Blender.

(I see the current version of Korman has a slightly different lightmapping modifier - I tried to replicate the issue on this version, but couldn't get a test Age to export at all, so sorry if this multi-user material bug has already been fixed :P )

Hoikas:
This should fix #132 and some other issues I observed when looking through the code.

Previously, we unconditionally exported materials. This worked in the situation that every material was single user. Now, the code should not assume that materials are single user and search for existing objects before creating them. If an object is lightmapped, Korman will now force it to have its own material. If you're sharing a non-lightmapped material, a lot of the material exporter code is now skipped. There is a bit of inefficiency when materials are shared among lightmapped objects in that the same layers will be converted multiple times. This is mostly just a time waster though and not an actual problem.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -178,9 +178,23 @@ class MaterialConverter:
 
     def export_material(self, bo, bm):
         """Exports a Blender Material as an hsGMaterial"""
-        self._report.msg("Exporting Material '{}'", bm.name, indent=1)
 
-        hsgmat = self._mgr.add_object(hsGMaterial, name=bm.name, bl=bo)
+        # Sometimes, a material might need to be single-use. Right now, the most apparent example
+        # of that situation is when a lightmap image is baked. Wavesets are in the same boat, but
+        # that's a special case as of the writing of this code.
+        single_user = self._requires_single_user_material(bo, bm)
+        if single_user:
+            mat_name = "{}_{}".format(bo.name, bm.name)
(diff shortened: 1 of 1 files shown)


Jrius:
That was fast ;)
I have a few issues on my end exporting with the current Korman (mostly an outdated HSPlasma but also a missing leading underscore for `_scale_image` in `texture.py` causing some errors), but with a few hacky edits I managed to get it to export a very basic scene. Looks ok, both lightmapped materials are now correctly duplicated.

Hoikas:
Life should become a bit easier for you when Korman 0.08 is released this week with an updated set of libraries

Deledrius:
@Jrius You'll need an updated PyHSPlasma and _korlib to use the current master.

Jrius:
> You'll need an updated PyHSPlasma and _korlib to use the current master

Yeah, I figured as much. For now I can work with Korman 0.07, although I'm looking forward to 0.08 :P

PR #161, opened 2019-10-28
Fix interesting empty node socket spawning bugs.

Hoikas:
This fixes two bugs with empty node socket spawning:
- The first time an empty is spawned, the link is moved to the last socket.
- Unlinking a socket in the middle of the group would cause that socket to be the one that's empty. This isn't bad per-se, but with Responders being ID-based, it could result in some confusion.

Beware: due to code movage, the diff looks strange.


Deledrius:
Animation Stages and Responder States can both be referenced by ID in scripts.  I'll have to think about this.

Hoikas:
Just looking to make sure it works correctly :wink:

Issue #155, opened 2019-10-23
Footstep and Ripple Decals


DoobesURU:

Hoikas:
Without going into the ability to add particles (yes, apparently you can do particles here...) we need the following bindings:
- [ ] `plDynaDecalMgr`
- [ ] `plDynaRippleMgr` - for rippling water due to non-feet
- [ ] `plDynaRippleVSMgr` (use in wavesets)
- [ ] `plDynaPuddleMgr` - for rippling water due to feet

PR #164, opened 2019-11-08
Add Blend Texture Exporter

Hoikas:
This adds support for the BlendTexture type to the material exporter. This has been done in preparation for #155, which will need alpha blend layers to be generated for certain footprint materials. The math was, in some cases, lifted from PyPRP.


Deledrius:
Looks good, works great!

PR #172, opened 2019-12-21
Decals

### LINKED ISSUE #155: Footstep and Ripple Decals

Hoikas:
This changeset adds dynamic decals and helps facilitate static, artist created decals.

In my investigation of static decals in the PlasmaMAX plugin, I determined that zflags only need to be applied to the material's base layer and that parenting is just a methodology used by Cyan--potentially to prevent decals from being separated from their objects by errant moving--that has no particular effect WRT decals at runtime or export time.

Note that dynamic decals are bugged in the MOUL engine. You can observe this by comparing footprints in Ahnonay Sphere 1 from PotS to MOUL. In PotS, the avatar causes water ripples and prints wet footprints on the ground after leaving the water. In MOUL, these do not appear to function although the objects are set up in an indistinguishable way. In my observations, only the alpha blending mode works reliably in MOUL.

This supersedes and closes #170. Fixes #155.

### DIFF
--- a/korman/exporter/decal.py
+++ b/korman/exporter/decal.py
@@ -15,6 +15,7 @@
 
 import bpy
 from collections import defaultdict
+import itertools
 from PyHSPlasma import *
 import weakref
 
@@ -34,7 +35,8 @@ def _get_footprint_class(exporter, name, vs):
 
 class DecalConverter:
     _decal_lookup = {
-        "footprint": _get_footprint_class,
+        "footprint_dry": _get_footprint_class,
+        "footprint_wet": _get_footprint_class,
         "puddle": _get_puddle_class,
         "ripple": lambda e, name, vs: plDynaRippleVSMgr if vs else plDynaRippleMgr,
     }
@@ -42,6 +44,7 @@ class DecalConverter:
     def __init__(self, exporter):
         self._decal_managers = defaultdict(list)
         self._exporter = weakref.ref(exporter)
+        self._notifies = defaultdict(set)
 
     def add_dynamic_decal_receiver(self, so, decal_name):
         # One decal manager in Blender can map to many Plasma decal managers.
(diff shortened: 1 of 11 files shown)


korman/exporter/decal.py
```diff
@@ -0,0 +1,174 @@
...
+            if decal_type in {"puddle", "ripple"}:
+                decal_mgr.wetLength = decal.wet_time
+                self._notifies[decal_name].extend((i.name for i in decal.wet_managers
+                                                          if i.enabled))
+
+            # UV Animations are hardcoded in PlasmaMAX. Any reason why we should expose this?
+            # I can't think of any presently... Note testing the final instance instead of the
+            # artist setting in case that gets overridden (puddle -> ripple)
+            if isinstance(decal_mgr, plDynaPuddleMgr):
+                decal_mgr.initUVW = hsVector3(0.5, 0.5, 0.5)
```
dpogue: These are actually [defaulted to 5.0](https://github.com/H-uru/Plasma/blob/master/Sources/Tools/MaxComponent/plFootPrintComponent.cpp#L872), not 0.5

korman/exporter/decal.py
```diff
@@ -0,0 +1,174 @@
...
+                decal_mgr.finalUVW = hsVector3(1.0, 1.0, 1.0)
+            elif isinstance(decal_mgr, plDynaRippleMgr):
+                # wakes, torpedos, and ripples...
+                decal_mgr.initUVW = hsVector3(0.3, 0.3, 0.3)
```
dpogue: Likewise, these are [defaulted to 3.0](https://github.com/H-uru/Plasma/blob/master/Sources/Tools/MaxComponent/plFootPrintComponent.cpp#L704), not 0.3

dpogue:
I think it's a bit weird conceptually that setting up print shapes (which cause decals) and setting up static decals are done using the same modifier. Not really an issue once you know it, but it's going to be super weird trying to explain that difference in documentation.

There is also a disappointing lack of torpedo support :stuck_out_tongue:

korman/exporter/decal.py
Hoikas: I was wondering why the ripple animations looked wrong! I thought I had inverted the init and final values, but, no I inverted the decimal. Thanks!

Hoikas:
My thinking was that they're fairly closely related since each case causes a decal to be printed onto a surface. To be honest, I don't expect for many folks to actually use the static decal portion, but I could be proven wrong.

While the exporter strictly speaking supports bullets, torpedoes, and wakes, it didn't seem particularly relevant to add support for them until particles are in. Even so, given that you pretty much have to fire bullets (and therefore torpedoes as well) manually through either nodes or Python, their use case seems more limited than the present offerings.

dpogue:
> My thinking was that they're fairly closely related since each case causes a decal to be printed onto a surface. To be honest, I don't expect for many folks to actually use the static decal portion, but I could be proven wrong.

I just found it non-intuitive (especially the name "print decal") but that probably isn't a big deal in the long run.

PR #176, opened 2020-02-16
Image Operator Menu

Hoikas:
Adds some hopefully useful operators for baking alpha to an image and clamping an image to on/off alpha (for forcing DXT1 export). Also fixes an issue with the cube map operator where the progress meter could run forever in the case of an error.


Deledrius:
This didn't work as expected for the tre'bivdil stained glass.  I'll try to provide a test case or debug it directly.

PR #180, opened 2020-03-25
Add a simple LinkingBook Modifier.

Deledrius:
Adds a simplified LinkingBook Modifier which automatically creates the PFM and associated nodes, as well as a temporary default region for the conditional if one is not provided by the artist.

### DIFF
--- a/korman/exporter/utils.py
+++ b/korman/exporter/utils.py
@@ -13,6 +13,11 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
+import bmesh
+import bpy
+from typing import Callable
+from contextlib import contextmanager
+
 from PyHSPlasma import *
 
 def affine_parts(xform):
@@ -45,3 +50,24 @@ def matrix44(blmat):
 def quaternion(blquat):
     """Converts a mathutils.Quaternion to an hsQuat"""
     return hsQuat(blquat.x, blquat.y, blquat.z, blquat.w)
+
+
+@contextmanager
+def bmesh_temporary_object(name : str, factory : Callable, page_name : str=None):
+    """Creates a temporary object and mesh that exists only for the duration of
+       the context"""
+    mesh = bpy.data.meshes.new(name)
+    obj = bpy.data.objects.new(name, mesh)
+    obj.draw_type = "WIRE"
+    if page_name is not None:
+        obj.plasma_object.page = page_name
+    bpy.context.scene.objects.link(obj)
+
+    bm = bmesh.new()
+    try:
+        factory(bm)
+        bm.to_mesh(mesh)
+        yield obj
+    finally:
+        bm.free()
+        bpy.context.scene.objects.unlink(obj)
--- a/korman/properties/modifiers/gui.py
+++ b/korman/properties/modifiers/gui.py
@@ -352,6 +352,7 @@ linking_pfms = {
             { 'id':  9, 'type': "ptAttribFloat",     "name": "bookWidth" },
             { 'id': 10, 'type': "ptAttribFloat",     "name": "BookHeight" },
             { 'id': 11, 'type': "ptAttribBehavior",  "name": "msbSeekBeforeUI" },
+            { 'id': 12, 'type': "ptAttribResponder", "name": "respOneShot" },
         )
     },
     pvMoul : {
@@ -583,6 +584,19 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         msb.link_input(seek, "seekers", "seek_target")
         msb.link_input(anim_stage, "stage", "stage_refs")
         msb.link_output(linkingnode, "hosts", "msbSeekBeforeUI")
+
+        # Responder
+        one_shot = nodes.new("PlasmaOneShotMsgNode")
+        one_shot.animation = self.anim_type
+        one_shot.marker = "touch"
+        one_shot.pos_object = self.seek_point
+
+        responder_state = nodes.new("PlasmaResponderStateNode")
+        responder_state.link_output(one_shot, "msgs", "sender")
+
+        responder = nodes.new("PlasmaResponderNode")
+        responder.link_output(responder_state, "state_refs", "resp")
+        responder.link_output(linkingnode, "keyref", "respOneShot")
 
     def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region):
         # Clickable
--- a/korman/properties/modifiers/gui.py
+++ b/korman/properties/modifiers/gui.py
@@ -17,12 +17,15 @@ import bpy
 import math
 import bmesh
 import mathutils
+from pathlib import Path
 from bpy.props import *
 from PyHSPlasma import *
 
 from ...addon_prefs import game_versions
+from ...exporter import ExportError, utils
 from .base import PlasmaModifierProperties, PlasmaModifierLogicWiz, PlasmaModifierUpgradable
 from ... import idprops
+from ...helpers import TemporaryObject
 
 
 journal_pfms = {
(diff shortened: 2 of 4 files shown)


Deledrius:
This feature still needs to have support for Path of the Shell exports added before it can be considered complete.

korman/ui/modifiers/gui.py
```diff
@@ -67,3 +67,58 @@ def journalbookmod(modifier, layout, context):
...
+            row.alert = not modifier.age_parent
+            row.prop(modifier, "age_parent")
+        if modifier.link_type == "kBasicLink":
+            row = layout.row()
+            row.alert = not modifier.age_uuid
+            row.prop(modifier, "age_uuid")
+
+    row = layout.row()
+    row.alert = not modifier.age_name
+    row.prop(modifier, "age_name")
```
Hoikas, line 102: Looks like this same antipattern appears 6 times. Maybe break it out into a function?
```python
def linkingbookmod(...):
    def row_alert(prop_name, **kwargs):
        row = layout.row()
        row.alert = not getattr(modifier, prop_name)
        row.prop(modifier, prop_name, **kwargs)
    ...
```

korman/exporter/utils.py
```diff
@@ -45,3 +50,22 @@ def matrix44(blmat):
 def quaternion(blquat):
     """Converts a mathutils.Quaternion to an hsQuat"""
     return hsQuat(blquat.x, blquat.y, blquat.z, blquat.w)
+
+
+@contextmanager
+def bmesh_temporary_object(name : str, factory : Callable, page_name : str=None):
```
Hoikas, line 56: A docstring would be nice ^_^

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,302 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+    stamped_book = BoolProperty(name="Has Stamp",
+                                description="Enables a stamp on the page opposite the book's linking panel",
+                                default=False)
+    stamp_image = PointerProperty(name="Stamp Image",
+                                  description="Image to use for the stamp",
+                                  type=bpy.types.Image,
+                                  options=set())
+    stamp_x = IntProperty(name="Stamp Position X",
+                          description="Width scale",
```
Hoikas, line 459: Scale?

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,302 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+                          default=140,
+                          subtype="UNSIGNED")
+    stamp_y = IntProperty(name="Stamp Position Y",
+                          description="Height scale",
```
Hoikas, line 463: Scale?

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,302 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+                                 bo.name, version, indent=2)
+            return
+
+        if self.clickable is None:
+            raise ExportError("Linking Book modifier requires a clickable!", bo.name)
+
+        if self.seek_point is None:
+            raise ExportError("Linking Book modifier requires a seek point!", bo.name)
+
+        if version == pvPots:
```
Hoikas, line 482: Even though this modifier doesn't support ABM, I would feel better if this were `if version <= pvPots:`

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,302 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+            for image in (self.book_cover_image, self.link_panel_image, self.stamp_image):
+                exporter.mesh.material.export_prepared_image(owner=ilmod, image=image, allowed_formats={"JPG", "PNG"}, extension="hsm")
+
+        # Auto-generate a six-foot cube region around the clickable if none was provided.
+        if self.clickable_region is None:
+            # Create a region for the clickable's condition
+            def _make_rgn(bm):
+                bmesh.ops.create_cube(bm, size=(6.0))
+                bmesh.ops.transform(bm, matrix=mathutils.Matrix.Translation(self.clickable.location - bo.location), space=bo.matrix_world, verts=bm.verts)
```
Hoikas, line 494: This line makes me have to side scroll.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,302 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+            with utils.bmesh_temporary_object("{}_LinkingBook_ClkRgn".format(self.key_name), _make_rgn, self.clickable.plasma_object.page) as temp_rgn:
```
Hoikas, line 496: This line makes me have to side scroll.

korman/properties/modifiers/gui.py
Deledrius: Whoops!  Copied and only partially edited!

korman/properties/modifiers/gui.py
Deledrius: Don't you have an ultrawide screen?

korman/ui/modifiers/gui.py
Deledrius: This was one of several repetitive patterns I was noticing.  I know I mentioned at least one in IRC.

korman/exporter/utils.py
```diff
@@ -45,3 +50,24 @@ def matrix44(blmat):
...
+    """Creates a temporary object and mesh that exists only for the duration of
+    the context"""
```
Hoikas, line 58: spacing

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,316 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+                          default=255,
+                          subtype="UNSIGNED")
+
+    def export(self, exporter, bo, so):
+        our_versions = (globals()[j] for j in self.versions)
+        version = exporter.mgr.getVer()
+        if version not in our_versions:
+            # We aren't needed here
+            exporter.report.port("Object '{}' has a LinkingBookMod not enabled for export to the selected engine.  Skipping.",
+                                 bo.name, version, indent=2)
```
Hoikas, line 463: This will raise an exception due to too many format arguments. I would probably leave out `version` because it's an integer.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,316 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+            return
+
+        if self.clickable is None:
+            raise ExportError("Linking Book modifier requires a clickable!", bo.name)
```
Hoikas, line 467: No format parameter in the string.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,316 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+        if self.seek_point is None:
+            raise ExportError("Linking Book modifier requires a seek point!", bo.name)
```
Hoikas, line 470: No format parameter in the string.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,316 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+        if version <= pvPots:
+            # Create ImageLibraryMod in which to store the Cover, Linking Panel, and Stamp images
+            ilmod = exporter.mgr.find_create_object(plImageLibMod, so=so, name=self.key_name)
+
+            for image in (self.book_cover_image, self.link_panel_image, self.stamp_image):
+                if image:
```
Hoikas, line 477: Maybe use `itertools.takewhile` to make this a one-liner (that is perhaps broken up on multiple lines :trollface:)?

korman/properties/modifiers/gui.py
Deledrius: ...
> No need to format an export error.
You told me to remove them...

korman/properties/modifiers/gui.py
Deledrius: Ah, didn't know about that one.  I looked into a few one-liner options, but they didn't seem more readable.

korman/properties/modifiers/gui.py
Hoikas: I meant to remove the call to `format()`. It needs to be:
```python
raise ExportError("{}: Linking Book modifier requires a clickable!", bo.name)
```

Deledrius: Alright.  I assumed from the previous context that it would prepend the object name.  Gotcha.

korman/properties/modifiers/gui.py
Deledrius: Do we have an accessible method to get the game version name from a version integer value?

Hoikas: I don't think so

korman/properties/modifiers/gui.py
Hoikas: I think I generally do stuff like:
```python
for i in (j for j in (thing1, thing2) if j):
```

But this might be better:
```python
for i in itertools.takewhile(lambda x: x, (thing1, thing2)):
```

Deledrius: `takewhile` would be a poor fit, as it stops the moment it finds an item that doesn't suit it.  How about `[x for x in (self.book_cover_image, self.link_panel_image, self.stamp_image) if x is not None]`?

Deledrius: Obviously, I can previously assign that tuple to a name to make it all more readable.

Hoikas: That will work, just make it a generator expression by changing the braces to parens.

Deledrius: So, perhaps:

```
user_images = (self.book_cover_image, self.link_panel_image, self.stamp_image)
for image in (x for x in user_images if x is not None):
   ...
```

Hoikas:
```python
user_images = (i for i in (self.book_cover_image, self.link_panel_image, self.stamp_image)
                 if i is not None)
for i in user_images:
    ...
```

Deledrius: That's even better.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,319 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+                            options={"ENUM_FLAG"},
+                            default={"pvMoul"})
+
+    # Link Info
+    link_type = EnumProperty(name="Linking Type",
+                             description="The type of Link this Linking Book will use",
+                             items=[
+                                 ("kBasicLink", "Public Link", "Links to a public instance of the specified Age"),
+                                 ("kOriginalBook", "Private Link", "Links to a new or existing private instance of the specified Age"),
+                                 ("kSubAge", "Closed Loop Link", "Links between instances of the specifed Age and the current one"),],
```
Hoikas, line 395: `kSubAgeBook`

Deledrius: I suspect this was a victim of our changes from explicit to simplified options.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +333,321 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+        responder.link_output(responder_state, "state_refs", "resp")
+        responder.link_output(linkingnode, "keyref", "respLinkResponder")
+
+        # Linking Panel Name
+        linking_panel_name = nodes.new("PlasmaAttribStringNode")
+        linking_panel_name.value = self.age_name
+        linking_panel_name.link_output(linkingnode, "pfm", "TargetAge")
+
+    def harvest_actors(self):
+        return(self.clickable.name, self.seek_point.name)
```
Hoikas, line 653: AttributeError: NoneType does not have attribute name

Also, spacing

PR #181, opened 2020-03-25
Add Linking Book guide.

Deledrius:
This adds a new item to the `Add...Plasma...` menu which creates a set of pre-configured Linking Book objects.  It creates a guide the artist can line up with the linking panel on a linking book, along with a seek point and clickable region parented to the panel guide, as well as an enabled Linking Book modifier.  The seek point is placed at a pre-computed position so that it lines up with the linking panel guide properly for both the standing and kneeling Link Animations.

### DIFF
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -140,11 +140,9 @@ class PlasmaAddLadderMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
             row.label("Warning: Operator does not work in local view mode", icon="ERROR")
 
     def execute(self, context):
-        if context.mode == "OBJECT":
-            self.create_ladder_objects()
-        else:
-            self.report({"WARNING"}, "Ladder creation only valid in Object mode")
-            return {"CANCELLED"}
+        if context.space_data.local_view:
+            bpy.ops.view3d.localview()
+        self.create_ladder_objects()
         return {"FINISHED"}
 
     def create_guide_rungs(self):
@@ -431,13 +429,10 @@ class PlasmaAddLinkingBookMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
             row.label("Warning: Operator does not work in local view mode", icon="ERROR")
 
     def execute(self, context):
-        if context.mode == "OBJECT":
-            self.create_linkingbook_objects()
-        else:
-            self.report({"WARNING"}, "Linking Book creation only valid in Object mode")
-            return {"CANCELLED"}
+        if context.space_data.local_view:
+            bpy.ops.view3d.localview()
+        self.create_linkingbook_objects()
         return {"FINISHED"}
-
 
     def create_linkingbook_objects(self):
         bpyscene = bpy.context.scene
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -18,10 +18,12 @@ import bmesh
 import math
 import mathutils
 
+from ..exporter import utils
+
 class PlasmaMeshOperator:
     @classmethod
     def poll(cls, context):
-        return context.scene.render.engine == "PLASMA_GAME"
+        return context.scene.render.engine == "PLASMA_GAME" and context.mode == "OBJECT"
 
 
 class PlasmaAddLadderMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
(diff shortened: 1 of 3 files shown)


korman/operators/op_mesh.py
```diff
@@ -387,6 +387,120 @@ def origin_to_bottom(obj):
...
+
+    # Allows user to specify their own name stem
+    panel_name = bpy.props.StringProperty(name="Name",
+                                         description="Linking Book name stem",
+                                         default="LinkingBook",
+                                         options=set())
+    link_anim_type = bpy.props.EnumProperty(name="Link Animation",
+                                            description="Type of Linking Animation to use",
+                                            items=[("Stand", "Standing", "The avatar steps up to the book and places their hand on the panel"),
+                                                   ("Kneel", "Kneeling", "The avatar kneels in front of the book and places their hand on the panel"),],
```
Hoikas: Maybe it would be better to use the animation names: `LinkOut`, `FishBookLinkOut` as the keys.

korman/operators/op_mesh.py
```diff
@@ -387,6 +387,120 @@ def origin_to_bottom(obj):
...
+            row.prop(self, "link_anim_type", text="Type")
+        else:
+            row = layout.row()
+            row.label("Warning: Operator does not work in local view mode", icon="ERROR")
+
+    def execute(self, context):
+        if context.mode == "OBJECT":
+            self.create_linkingbook_objects()
+        else:
+            self.report({"WARNING"}, "Linking Book creation only valid in Object mode")
```
Hoikas: Add this check to the operator `poll()` for less error checking logic.

korman/operators/op_mesh.py
```diff
@@ -387,6 +387,120 @@ def origin_to_bottom(obj):
...
+        # Add the object into the scene.
+        bpyscene.objects.link(clk_rgn)
+        clk_rgn.plasma_object.enabled = True
+
+        # Construct the clickable region and assign it to the blender mesh.
+        bm = bmesh.new()
+        clk_rgn_size = 6.0
+        bmesh.ops.create_cube(bm, size=(1.0), matrix=(mathutils.Matrix.Scale(clk_rgn_size, 4)))
+        bm.to_mesh(clk_rgn_mesh)
+        bm.free()
```
Hoikas: I've seen this same anti-pattern in two PRs now. Maybe it would be best to write a contextlib helper so we can do something like:
```python
with utils.bmesh_object("My Fancy Object") as clk_rgn, bm:
        clk_rgn_size = 6.0
        bmesh.ops.create_cube(bm, size=(1.0), matrix=(mathutils.Matrix.Scale(clk_rgn_size, 4)))
```

Implementation would look something like pseudo-python:
```python
@contextmanager
def bmesh_object(name):
    mesh = bpy.data.meshes.new(name)
    obj = bpy.data.objects.new(name, mesh)
    obj.draw_type = "WIRE"
    bpy.context.scene.objects.link(obj)
    bm = bmesh.new()
    try:
        yield obj, bm
    else:
        bm.to_mesh(mesh)
    finally:
        bm.free()
```

korman/operators/op_mesh.py
```diff
@@ -387,6 +387,120 @@ def origin_to_bottom(obj):
...
+
+        # Parent Region and SeekPoint to Panel
+        seek_point.parent = panel_root
+        clk_rgn.parent = panel_root
+
+        # Add Linking Book modifier
+        bpyscene.objects.active = panel_root
+        panel_root.select = True
+        bpy.ops.object.plasma_modifier_add(types="linkingbookmod")
+        panel_root.plasma_modifiers.linkingbookmod.clickable_region = clk_rgn
```
Hoikas: Readability: make an alias for plasma_modifiers

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +331,249 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+        version = exporter.mgr.getVer()
+        if version not in our_versions:
+            # We aren't needed here
+            exporter.report.port("Object '{}' has a LinkingBookMod not enabled for export to the selected engine.  Skipping.",
+                                 bo.name, version, indent=2)
+            return
+
+
+        if self.clickable is None:
+            raise ExportError("{}: Linking Book modifier requires a clickable!".format(bo.name))
```
Hoikas: No need to format an export error.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +331,249 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+
+        if self.seek_point is None:
+            raise ExportError("{}: Linking Book modifier requires a seek point!".format(bo.name))
```
Hoikas: No need to format an export error.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +331,249 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+        # Auto-generate a six-foot cube region around the clickable if none was provided.
+        if self.clickable_region is None:
+            # Create a region for the clickable's condition
+            rgn_mesh = bpy.data.meshes.new("{}_LinkingBook_ClkRgn".format(self.key_name))
+            self.temp_rgn = bpy.data.objects.new("{}_LinkingBook_ClkRgn".format(self.key_name), rgn_mesh)
+            bm = bmesh.new()
+            bmesh.ops.create_cube(bm, size=(6.0))
+            bmesh.ops.transform(bm, matrix=mathutils.Matrix.Translation(self.clickable.location - self.temp_rgn.location), space=self.temp_rgn.matrix_world, verts=bm.verts)
+            bm.to_mesh(rgn_mesh)
+            bm.free()
```
Hoikas: See above re-antipattern.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +331,249 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+        with self.generate_logic(bo, age_name=exporter.age_name, version=version) as tree:
+            tree.export(exporter, bo, so)
+
+        # Get rid of our temporary clickable region
+        if self.clickable_region is None:
+            bpy.context.scene.objects.unlink(self.temp_rgn)
+
+        # Get rid of our temporary seek point
+        if self.clickable_region is None:
+            bpy.context.scene.objects.unlink(self.temp_seek)
```
Hoikas: If the exporter crashes, these objects are left orphaned. See `utils.TemporaryObject`.

korman/properties/modifiers/gui.py
```diff
@@ -330,3 +331,249 @@ def _create_moul_nodes(self, clickable_object, nodes, journalnode, age_name):
...
+                new_attr.attribute_name = attr["name"]
+        linkingnode.update()
+
+        if version <= pvPots:
+            self._create_pots_nodes(bo, nodes, linkingnode, age_name)
+        else:
+            self._create_moul_nodes(bo, nodes, linkingnode, age_name)
+
+    def _create_pots_nodes(self, clickable_object, nodes, linkingnode, age_name):
+        pass
```
Hoikas: Do not pass go. Do not collect nodes. :trollface:

korman/operators/op_mesh.py
Deledrius: It's used in more places than that!
I welcome a better solution, so this looks good to me.

korman/properties/modifiers/gui.py
Deledrius: Except that it's really useful having the error actually tell you which object caused the export failure.

korman/properties/modifiers/gui.py
Deledrius: Oh, yeah!  Forgot you added that!

korman/properties/modifiers/gui.py
Deledrius: Not yet, anyway, hence the comment on #180, the PR in which this code was submitted. ;)

korman/properties/modifiers/gui.py
Hoikas: Clarification: pass `bo.name` as an argument to `ExportError`--it applies `format()` for you.

Deledrius: Ah, I see!

korman/operators/op_mesh.py
Deledrius: Done.  However, it now fails ungracefully if one tries to Add while in Local View mode (which should be disabled).  I'm investigating the correct way to filter this.

Deledrius: Looks like a good solution is to toggle out of Local View mode if we're in it before executing the function here.

```
if context.space_data.local_view:
    bpy.ops.view3d.localview()
```

korman/operators/op_mesh.py
```diff
@@ -18,10 +18,15 @@
...
+from ..exporter import utils
+
 class PlasmaMeshOperator:
     @classmethod
     def poll(cls, context):
-        return context.scene.render.engine == "PLASMA_GAME"
+        if context.mode != "OBJECT":
+            return False
+        else:
+            return context.scene.render.engine == "PLASMA_GAME"
```
Hoikas, line 29:
```python
return context.scene.render.engine == "PLASMA_GAME" and context.mode == "OBJECT"
```

Deledrius: Sure, if you want to do it properly...

Hoikas: :trollface:

PR #182, opened 2020-03-25
Add premade Lamp Flare.

Deledrius:
This adds a new item to the `Add...Plasma...` menu which provides a pre-configured lamp flare.  

It creates a set of objects for an optimal, general-purpose glow around most lamps:
- An empty with a Swivel modifier at the chosen location (usually the lamp).
- A plane parented to the empty for the flare texture, with an Opacity Fade modifier which uses the plane's center for Line-of-Sight.

### DIFF
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -26,6 +26,143 @@ class PlasmaMeshOperator:
         return context.scene.render.engine == "PLASMA_GAME" and context.mode == "OBJECT"
 
 
+FLARE_MATERIAL_BASE_NAME = "FLAREGEN"
+
+def store_material_selection(self, value):
+    if bpy.data.materials.get(value, None):
+        bpy.context.scene.plasma_scene.last_flare_material = value
+
+
+class PlasmaAddFlareOperator(PlasmaMeshOperator, bpy.types.Operator):
+    bl_idname = "mesh.plasma_flare_add"
+    bl_label = "Add Lamp Flare"
+    bl_category = "Plasma"
+    bl_description = "Adds a new Plasma Lamp Flare"
+    bl_options = {"REGISTER", "UNDO"}
+
+    # Allows user to specify their own name stem
+    flare_name = bpy.props.StringProperty(name="Name",
+                                         description="Flare name stem",
+                                         default="Flare",
+                                         options=set())
+    flare_distance = bpy.props.FloatProperty(name="Distance",
+                                             description="Flare's distance from the illuminating object",
+                                             min=0.1, max=2.0, step=10, precision=1, default=1.0,
+                                             options=set())
+    flare_material_name = bpy.props.StringProperty(name="Material",
+                                                   description="A specially-crafted material to use for this flare",
+                                                   default=FLARE_MATERIAL_BASE_NAME,
+                                                   update=store_material_selection,
+                                                   options=set())
+
+    @classmethod
+    def poll(cls, context):
+        return super().poll(context) and context.mode == "OBJECT"
+
+    def draw(self, context):
+        layout = self.layout
+        box = layout.box()
+        box.label("Flare Name:")
+        row = box.row()
+        row.alert = not self.flare_name
+        row.prop(self, "flare_name", text="")
+
+        box = layout.box()
+        box.label("Geometry:")
+        row = box.row()
+        row.prop(self, "flare_distance")
+
+        box = layout.box()
+        box.label("Material:")
+        row = box.row()
+        row.prop_search(self, "flare_material_name", bpy.data, "materials", text="")
+
+    def execute(self, context):
+        if context.space_data.local_view:
+            bpy.ops.view3d.localview()
+        self.create_flare_objects()
+        return {"FINISHED"}
+
+    def create_flare_objects(self):
+        bpyscene = bpy.context.scene
+        cursor_shift = mathutils.Matrix.Translation(bpy.context.scene.cursor_location)
+
+        for obj in bpy.data.objects:
+            obj.select = False
+
+        # Create Empty, rotated to horizontal
+        flare_root = bpy.data.objects.new("{}".format(self.name_stem), None)
+        flare_root.empty_draw_type = "SINGLE_ARROW"
+        flare_root.matrix_world = cursor_shift
+        flare_root.rotation_euler[0] = math.radians(90)
+        flare_root.plasma_object.enabled = True
+        flare_root.select = True
+        bpy.context.scene.objects.link(flare_root)
+        bpyscene.objects.active = flare_root
+
+        # Enable VFM on Empty
+        bpy.ops.object.plasma_modifier_add(types="viewfacemod")
+        flare_root.plasma_modifiers.viewfacemod.preset_options = "Sprite"
+
+        # Create a textured Plane
+        with utils.bmesh_object("{}_Visible".format(self.name_stem)) as (flare_plane, bm):
+            flare_plane.hide_render = True
+            flare_plane.plasma_object.enabled = True
+            bpyscene.objects.active = flare_plane
+
+            # Make the actual plane mesh, facing away from the empty
+            bmesh.ops.create_grid(bm, size=(0.5 + self.flare_distance * 0.5), matrix=mathutils.Matrix.Rotation(math.radians(180.0), 4, 'X'))
+            bmesh.ops.transform(bm, matrix=mathutils.Matrix.Translation((0.0, 0.0, -self.flare_distance)), space=flare_plane.matrix_world, verts=bm.verts)
+            bpy.ops.object.origin_set(type="ORIGIN_GEOMETRY")
+
+        # Give the plane a basic UV unwrap, so that it's texture-ready
+        bpy.ops.object.editmode_toggle()
+        bpy.ops.uv.smart_project()
+        bpy.ops.object.editmode_toggle()
+
+        # Create AUTOGEN_FLARE material and texture, setting No-Z-Write
+        auto_mat = self.find_create_material()
+        flare_plane.data.materials.append(auto_mat)
+
+        # Parent Plane to Empty
+        flare_plane.parent = flare_root
+
+        # Enable Opacity Fader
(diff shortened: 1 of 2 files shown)


Deledrius:
As noted in the WIP comment, still missing is the necessary Plasma Layer options, which would need to be set on a texture.  I'm submitting this as a draft PR to solicit suggestions on how best to accomplish this.  Should we pregenerate a default material and texture (and reuse it for subsequent flares), setting the flags needed such as No Z-Write?

korman/operators/op_mesh.py
```diff
@@ -24,6 +24,109 @@ def poll(cls, context):
...
+            row.prop(self, "flare_distance")
+        else:
+            row = layout.row()
+            row.label("Warning: Operator does not work in local view mode", icon="ERROR")
+
+    def execute(self, context):
+        if context.mode == "OBJECT":
+            self.create_flare_objects()
+        else:
+            self.report({"WARNING"}, "Lamp Flare creation only valid in Object mode")
```
Hoikas: Might be better to adjust the `poll()` to check for object mode.

korman/operators/op_mesh.py
```diff
@@ -24,6 +24,109 @@ def poll(cls, context):
...
+        cursor_shift = mathutils.Matrix.Translation(bpy.context.scene.cursor_location)
+
+        for obj in bpy.data.objects:
+            obj.select = False
+
+        # Create Empty, rotated to horizontal
+        flare_root = bpy.data.objects.new("{}".format(self.name_stem), None)
+        flare_root.empty_draw_type = "SINGLE_ARROW"
+        flare_root.matrix_world = cursor_shift
+        flare_root.rotation_euler[0] = math.pi/2
```
Hoikas: Style nit

korman/ui/ui_menus.py
```diff
@@ -30,6 +30,7 @@ def draw(self, context):
         layout = self.layout
 
         layout.operator("mesh.plasma_ladder_add", text="Ladder", icon="COLLAPSEMENU")
+        layout.operator("mesh.plasma_flare_add", text="Lens Flare", icon="PARTICLES")
```
Hoikas: Lamp Flare might be better? I would say a lens flare is a camera effect.

Deledrius: Whoops, missed this in the rename.

korman/operators/op_mesh.py
Deledrius: 👍

Deledrius:
There isn't an obvious way to do that for a Mesh Operator, though.  Should I have it autogen a Material and select it, with a material selection box in the operator?  That allows us to set up the plLayer settings, but it'll still need an image assigned to the texture slot.

The ideal workflow is as follows: artist places the 3D cursor on their lamp mesh, presses Shift-A to Add->Plasma->Lamp Flare, and is given the current operator with an option to name it and the distance from the lamp to the flare so they can adjust it in real time before committing.  It's parented, so if the size of the mesh changes or adjustments are needed, they can move the flare mesh on the Y-axis without breaking anything.

What's missing is the material.  An operator can't easily *require* a field, so if it's possible to leave either an image or material selection blank, the flare will be in an unusable state that isn't easily resolved without a walkthrough (which defeats the entire purpose of these).  There's nothing "special" being created here, just pre-assembling a bunch of existing parts and modifiers using the settings we know need to be set behind-the-scenes.  This leads me to think we should auto-generate on first-use a flare material with the appropriate settings.  Unless we provide an image texture, this will still need to be set in some way.  On subsequent uses, we can auto-select the last material matching the pre-gen name, or allow creation of a new one (for a different image/color tint flare).  I'm not sure that there's any way to flag to the user that a flare is missing its image and guide them to rectifying this.

Deledrius:
Note: this contains some duplicate code from #181.  Once that's merged I'll rebase and remove the redundant bits here, and then squash these two commits.

korman/operators/op_mesh.py
```diff
@@ -18,12 +18,133 @@
...
+        cursor_shift = mathutils.Matrix.Translation(bpy.context.scene.cursor_location)
+
+        for obj in bpy.data.objects:
+            obj.select = False
+
+        # Create Empty, rotated to horizontal
+        flare_root = bpy.data.objects.new("{}".format(self.name_stem), None)
+        flare_root.empty_draw_type = "SINGLE_ARROW"
+        flare_root.matrix_world = cursor_shift
+        flare_root.rotation_euler[0] = math.pi / 2
```
Hoikas, line 80: Why not `math.radians(90)`? Requires less thinking on my part :)

korman/operators/op_mesh.py
```diff
@@ -18,12 +18,133 @@
...
+        flare_plane.plasma_modifiers.fademod.fade_out_time = 0.25
+        flare_plane.plasma_modifiers.fademod.bounds_center = True
+
+    def find_create_material(self):
+        material_base_name = "AUTOGEN_FLARE"
+        # If one or more auto-generated flare materials already exist, use the newest one
+        matlist = [mat for mat in bpy.data.materials if mat.name.startswith(material_base_name)]
+        if matlist:
+            matlist.sort()
+            return matlist[-1]
```
Hoikas, line 126: I feel like there is a better way to do this.

korman/operators/op_mesh.py
```diff
@@ -26,6 +26,13 @@ def poll(cls, context):
         return context.scene.render.engine == "PLASMA_GAME" and context.mode == "OBJECT"
 
 
+FLARE_MATERIAL_BASE_NAME = "FLAREGEN"
+
+def _flare_materials(self, context):
+    matlist = [mat for mat in bpy.data.materials if mat.name.startswith(FLARE_MATERIAL_BASE_NAME)]
+    if matlist:
+        return [(mat.name, mat.name, "", "MATERIAL", idx) for idx, mat in enumerate(matlist)]
```
Hoikas, line 34: If at all possible, avoid including a numerical index in this enum property. In the best case, you're just going to see the odd behavior where either the getter or setter expects an `str` and the other expects an `int`. At worst, the ability of the operator to recall the previous selection is impaired due to Blender saving the index into matlist, which is potentially nondeterminant.

korman/operators/op_mesh.py
```diff
@@ -26,6 +26,13 @@ def poll(cls, context):
         return context.scene.render.engine == "PLASMA_GAME" and context.mode == "OBJECT"
 
 
+FLARE_MATERIAL_BASE_NAME = "FLAREGEN"
+
+def _flare_materials(self, context):
+    matlist = [mat for mat in bpy.data.materials if mat.name.startswith(FLARE_MATERIAL_BASE_NAME)]
```
Hoikas, line 32: It would be better to use a generator expression here.

korman/operators/op_mesh.py
```diff
@@ -42,6 +49,10 @@ class PlasmaAddFlareOperator(PlasmaMeshOperator, bpy.types.Operator):
                                              description="Flare's distance from the illuminating object",
                                              min=0.1, max=2.0, step=10, precision=1, default=1.0,
                                              options=set())
+    flare_material = bpy.props.EnumProperty(name="Material",
+                                            description="",
+                                            default=None,
```
Hoikas, line 54: Is this even legal? In blender 2.80, properties become type annotations, so I'm a little hesitant on this.

korman/operators/op_mesh.py
```diff
@@ -118,25 +134,23 @@ def create_flare_objects(self):
...
 
     def find_create_material(self):
-        material_base_name = "AUTOGEN_FLARE"
-        # If one or more auto-generated flare materials already exist, use the newest one
-        matlist = [mat for mat in bpy.data.materials if mat.name.startswith(material_base_name)]
-        if matlist:
-            matlist.sort()
-            return matlist[-1]
+        # If an auto-generated flare materials exists, use the selected one
+        if self.flare_material is not None and self.flare_material in bpy.data.materials.keys():
```
Hoikas, line 138: See above re type confusion on `None`. This could probably be simplified to:
```python
bpy.data.materials.get(self.fare_material, None)
```

korman/operators/op_mesh.py
```diff
@@ -118,27 +135,26 @@ def create_flare_objects(self):
...
-
-        auto_tex = bpy.data.textures.new(name=material_base_name, type="IMAGE")
-        auto_tex.use_alpha = True
-        auto_tex.plasma_layer.skip_depth_write = True
-
-        new_slot = auto_mat.texture_slots.add()
-        new_slot.texture = auto_tex
-        return auto_mat
+        # If the selected flare material exists, use it
+        auto_mat = bpy.data.materials.get(self.flare_material_name)
```
Hoikas, line 139: IIRC blender's `bpy_prop_collection` will raise an exception on error if a default is not provided.

korman/operators/op_mesh.py
Deledrius: I guess I naturally think in terms of rotations here instead of degrees, since that's the goal of this, so `pi / 2` made the most sense.  I'll change it though so it'll be consistent.

korman/operators/op_mesh.py
```diff
@@ -26,6 +26,141 @@ def poll(cls, context):
...
+            # Generate a new flare material and texture
+            auto_mat = bpy.data.materials.new(name=FLARE_MATERIAL_BASE_NAME)
+            auto_mat.use_shadeless = True
+            auto_mat.use_shadows = False
+            auto_mat.use_cast_shadows = False
+            self.flare_material_name = auto_mat.name
+
+            auto_tex = bpy.data.textures.new(name=FLARE_MATERIAL_BASE_NAME, type="IMAGE")
+            auto_tex.use_alpha = True
+            auto_tex.plasma_layer.skip_depth_write = True
```
dpogue, line 151: You probably also want to set `layer_props.z_bias` and `layer_props.skip_depth_test`. These allow the flare to shine past the edges of partially occluding objects (but then the FadeOpacityMod hides them when completely occluded)

Deledrius: Thanks, I'll try those out!

Deledrius: Added!

PR #186, opened 2020-06-30
Add missing descriptions for PlasmaAnimStageSettingsNode properties.

Deledrius:
This adds descriptions to the node's fields which were accidentally left out of #124.


Deledrius:
This is somewhat nit-picking, but when writing them I found it difficult to describe the `anim_stage_flags` in a way that was clear and concise while covering both usages.  Would it be preferred to create an `anim_stage_adv_flags` and `anim_stage_reg_flags` instead so that we can clearly describe one as "Advancing to the next" and "Regressing to the previous"?

Hoikas:
Probably - the current wording is a bit confusing.

Deledrius:
Alright, I'll do that.  I just wanted a second opinion. :)

PR #190, opened 2020-07-24
Add Warp Message node.

Deledrius:
Adds a Warp Message node for use in advanced logic trees.

This node requires an object position, and sends a message to warp there to the object specified by a `PlasmaSceneObjectMsgRcvrNode`.

### DIFF
--- a/korman/nodes/node_messages.py
+++ b/korman/nodes/node_messages.py
@@ -20,7 +20,7 @@ from PyHSPlasma import *
 
 from .node_core import *
 from ..properties.modifiers.region import footstep_surfaces, footstep_surface_ids
-from ..exporter import ExportError
+from ..exporter import ExportError, utils
 from .. import idprops
 
 class PlasmaMessageSocketBase(PlasmaNodeSocketBase):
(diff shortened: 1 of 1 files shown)


korman/nodes/node_messages.py
```diff
@@ -877,3 +877,35 @@ def convert_message(self, exporter, so):
...
+
+    pos_object = PointerProperty(name="Position",
+                                 description="Object defining the target position",
+                                 type=bpy.types.Object)
+
+    output_sockets = OrderedDict([
+        ("receivers", {
+            "text": "Send To",
+            "type": "PlasmaWarpMessageSocket",
+            "valid_link_sockets": {"PlasmaWarpMessageSocket", "PlasmaNodeSocketInputGeneral"},
```
Hoikas, line 895: `"PlasmaNodeSocketInputGeneral"`?

korman/nodes/node_messages.py
```diff
@@ -877,3 +877,35 @@ def convert_message(self, exporter, so):
...
+    ])
+
+    def draw_buttons(self, context, layout):
+        layout.prop(self, "pos_object")
+
+    def convert_message(self, exporter, so):
+        msg = plWarpMsg()
+        msg.BCastFlags |= plMessage.kNetPropagate
+        msg.warpFlags |= plWarpMsg.kFlushTransform
+        msg.transform = utils.matrix44(self.pos_object.matrix_local)
```
Hoikas, line 906: I think this would be more useful if `pos_object` were optional and defaulted to the pose of the object the tree is attached to.

Deledrius: That's a useful default.  I'll give it a try and see how that works in practice.

korman/nodes/node_messages.py
Hoikas: After the explanation in IRC, it might be better to use `valid_link_nodes` instead and specify the expected node type.

PR #191, opened 2020-07-29
Add Help menu for Korman.

Deledrius:
This adds a Help menu to the appropriate place in Blender.  It provides links to common documents of use to Age builders.

### DIFF
--- a/korman/ui/ui_menus.py
+++ b/korman/ui/ui_menus.py
@@ -40,8 +40,25 @@ def build_plasma_menu(self, context):
     self.layout.separator()
     self.layout.menu("menu.plasma_add", icon="URL")
 
+
+class PlasmaHelpMenu(PlasmaMenu, bpy.types.Menu):
+    bl_idname = "menu.plasma_help"
+    bl_label = "Korman..."
+
+    def draw(self, context):
+        layout = self.layout
+        layout.operator("wm.url_open", text="About Korman", icon="URL").url = "https://guildofwriters.org/wiki/Korman"
+        layout.operator("wm.url_open", text="Getting Started", icon="URL").url = "https://guildofwriters.org/wiki/Korman:Getting_Started"
+        layout.operator("wm.url_open", text="Tutorials", icon="URL").url = "https://guildofwriters.org/wiki/Category:Korman_Tutorials"
+
+
+def build_plasma_help_menu(self, context):
+    self.layout.menu("menu.plasma_help", text="Korman", icon="URL")
+
 def register():
     bpy.types.INFO_MT_add.append(build_plasma_menu)
+    bpy.types.INFO_MT_help.prepend(build_plasma_help_menu)
 
 def unregister():
     bpy.types.INFO_MT_add.remove(build_plasma_menu)
+    bpy.types.INFO_MT_help.remove(build_plasma_help_menu)


korman/ui/ui_menus.py
```diff
@@ -40,8 +40,27 @@ def build_plasma_menu(self, context):
...
+class korman_help_menu(bpy.types.Menu):
+    bl_idname = "menu.plasma_help"
+    bl_label = "Korman..."
+
+    def draw(self, context):
+        layout = self.layout
+
+        layout.operator("wm.url_open", text="About Korman", icon='URL').url="https://guildofwriters.org/wiki/Korman"
+        layout.operator("wm.url_open", text="Getting Started", icon='URL').url="https://guildofwriters.org/wiki/Korman:Getting_Started"
+        layout.operator("wm.url_open", text="Tutorials", icon='URL').url="https://guildofwriters.org/wiki/Category:Korman_Tutorials"
```
Hoikas, line 53: Whitespace nit. `str` should be double quoted `"`

Deledrius:
Additional page suggestions for links are welcome.  I think it would also be great to have a link to a page that allows the user to easily check for updates, but that would require adding a page to the website for this purpose.

korman/ui/ui_menus.py
```diff
@@ -40,8 +40,27 @@ def build_plasma_menu(self, context):
...
+
+
+def build_korman_help_menu(self, context):
+    self.layout.menu("menu.plasma_help", text="Korman", icon='URL')
```
Hoikas, line 57: `str` should be double quoted `"`

korman/ui/ui_menus.py
```diff
@@ -40,8 +40,27 @@ def build_plasma_menu(self, context):
...
+
+
```
Hoikas, line 59: According to PEP8, functions shall be followed by only one blank line.

korman/ui/ui_menus.py
Deledrius: Copypasta errors. >_<

korman/ui/ui_menus.py
```diff
@@ -40,8 +40,25 @@ def build_plasma_menu(self, context):
...
+class korman_help_menu(bpy.types.Menu):
+    bl_idname = "menu.plasma_help"
+    bl_label = "Korman..."
+
+    def draw(self, context):
+        layout = self.layout
+        layout.operator("wm.url_open", text="About Korman", icon="URL").url="https://guildofwriters.org/wiki/Korman"
+        layout.operator("wm.url_open", text="Getting Started", icon="URL").url="https://guildofwriters.org/wiki/Korman:Getting_Started"
+        layout.operator("wm.url_open", text="Tutorials", icon="URL").url="https://guildofwriters.org/wiki/Category:Korman_Tutorials"
```
Hoikas, line 52: Whitespace

Deledrius:
That depends!  Are there any other obvious links we should include?

Hoikas:
I can't think of anything else right now.

DoobesURU:
I say the simpler, the better. This should be enough.

Deledrius:
There are things I want to add, but they can be additions.  This is a good baseline.

korman/ui/ui_menus.py
```diff
@@ -40,8 +40,25 @@ def build_plasma_menu(self, context):
     self.layout.separator()
     self.layout.menu("menu.plasma_add", icon="URL")
 
+
+class korman_help_menu(bpy.types.Menu):
```
Hoikas, line 44: The class name is not consistent with PEP8, and we need to inherit from `PlasmaMenu`.

Deledrius:
I'd like to change `build_plasma_menu` to be `build_plasma_add_menu` for consistency's sake.  Should I piggy-back that on here with another commit, or open a whole PR just for that?

Hoikas (approved):

Hoikas:
Let's hold off until another PR on that. It would make more sense to me if those were a `build_menu` method on the menu classes, but meh.

PR #193, opened 2020-08-22
Add support for pushable objects

dpogue:
I assume these are going to be necessary if we want to add the ball back to the Watcher's Pub.

Note: This currently doesn't work engine-side in current Plasma.


korman/ui/modifiers/physics.py
```diff
@@ -34,6 +34,9 @@ def collision(modifier, layout, context):
     row = col.row()
     row.active = modifier.dynamic
     row.prop(modifier, "start_asleep")
+    row = col.row()
+    row.active = modifier.dynamic
```
Hoikas, line 38:
```suggestion
```

PR #196, opened 2020-09-07
Font Export

julyfortoday:
Added export / conversion function for turning font objects into mesh objects that can be exported to Plasma.

Since I'm using so many font objects, I found myself wanting Korman to just automatically export them, instead of manually converting and managing a bunch of mesh copies. Thought I would take a crack at it, and managed to get it working.

Uses operators (discouraged practice, I know, but I couldn't figure out how to do it using low level functions after spending more time than I wanted looking through old Blender Python API documentation)

Deselects everything, selects the current blender object, converts it, then hands it over to the mesh export function, keeps the original and cleans up by deleting the converted mesh after export.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -273,18 +273,9 @@ class Exporter:
             self.report.msg("No material(s) on the ObData, so no drawables", indent=1)
 
     def _export_font_blobj(self, so, bo):
-        self.report.msg("Converting font to mesh for export")
-        bpy.ops.object.select_all(action='DESELECT')
-        bpy.context.scene.objects.active = bo
-        bo.select = True
-        convertible = bpy.ops.object.convert.poll()
-        if convertible:
-            bpy.ops.object.convert(target='MESH', keep_original= True)
-            convertedFont = bpy.context.active_object
-            self._export_mesh_blobj(so, convertedFont)
-            bpy.ops.object.delete()
-        else:
-            self.report.msg("not convertible, skipping...")
+        self.animation.convert_object_animations(bo, so)
+        with utils.temporary_mesh_object(bo) as meshObj:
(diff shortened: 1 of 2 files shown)


Hoikas:
Thanks for the pull request!

There are quite a few places where the user selection gets clobbered and success is assumed, potentially leaving behind orphaned or junked objects. I'll look into what needs to happen for this to be safer.

> Uses operators (discouraged practice, I know, but I couldn't figure out how to do it using low level functions after spending more time than I wanted looking through old Blender Python API documentation)

Yeah, sometimes there's just no other way. I feel like I spend half my time trying to figure out ways to avoid object operators (they are slow!).

Hoikas:
@julyfortoday Can you verify that the simplification I did in e1b9537 works correctly for you?

julyfortoday:
@Hoikas I can confirm, it works (and faster than my hack does).

I understand some of your improved code is doing, but I will confess to having absolutely no clue how it actually does the font to mesh conversion, lol.

You have my gratitude for the simplification (and for your patience in dealing with Blender's Python API -- I always end up with a headache after trying to do something with it).

I'm not sure if I have to do anything on my end (to resolve the pull-request -- this is actually the first one I have done - Yay for new experiences! hehe )

Hoikas:
> You have my gratitude for the simplification (and for your patience in dealing with Blender's Python API -- I always end up with a headache after trying to do something with it).

I usually wind up experimenting heavily in the blender console and sometimes reading blender's source code itself - yuck.

> I'm not sure if I have to do anything on my end (to resolve the pull-request -- this is actually the first one I have done - Yay for new experiences! hehe )

Nothing further. I'm going to press the shiny green merge button after this comment. Thanks again for the pull request!

PR #197, opened 2020-09-08
Prototype: SharedMesh export

dpogue:
This _works_ but isn't useful right now, because we lack bone weights and morph sets and clothing items. But it does successfully export a plSharedMesh containing the right geometry, which is kinda cool.


Hoikas:
@Deledrius has thoughts about how to make all this clothing stuff make sense in the Blender context.

PR #200, opened 2020-09-14
Fix Rail Camera keyframe issue

DoobesURU:
Keeps Korman from exporting a rail camera with only one keyframe (or duplicate location keyframes), the former of which causes the client to crash.

### DIFF
--- a/korman/exporter/camera.py
+++ b/korman/exporter/camera.py
@@ -206,6 +206,9 @@ class CameraConverter:
             f1, f2 = fcurve.evaluate(begin), fcurve.evaluate(end)
             if abs(f1 - f2) > 0.001:
                 break
+            # to avoid single/duplicate keyframe client crash (per Hoikas)
+            if any((len(i.keys) == 1 for i in (pos_ctrl.X, pos_ctrl.Y, pos_ctrl.Z) if i is not None)):
+                raise ExportError("'{}': Rail Camera must have more than one keyframe", bo.name)
         else:
             # The animation is a loop
             path.flags |= plAnimPath.kWrap


korman/exporter/camera.py
```diff
@@ -206,6 +206,9 @@ def _export_rail_camera(self, so, bo, props):
             f1, f2 = fcurve.evaluate(begin), fcurve.evaluate(end)
             if abs(f1 - f2) > 0.001:
                 break
+            # to avoid single/duplicate keyframe client crash - Doobes
+            elif f1 == f2:
```
Hoikas, line 210: This will probably never get triggered. If I'm remembering correctly, it is legal for the first and last keyframe to be in roughly the same place, which triggers a special "wrapping" kind of rail camera. The if statement is to give some wiggle room for engaging that mode. I think what we need to do is make sure the final exported `pos_ctrl` has more than a single keyframe. We should probably be testing that deeper inside the animation code to make sure no animations have a single keyframe. For now, though, we can do something like:

```python
if any((len(i.keys == 1) for i in (pos_ctrl.X, pos_ctrl.Y, pos_ctrl.Z) if i is not None)):
    raise ExportError("'{}': Rail Camera must have more than one keyframe", bo.name)
```

before line 201

DoobesURU: If it's a wrapping camera, wouldn't a circle camera make more sense? Unless it's not a uniform circle path, I suppose.

Note that this *did* trigger in my testing of it, both with single keyframe and two keyframes in the same position. I suppose if the latter doesn't crash the client, then your code is probably much more the scalpel to my sledgehammer. 😉

Hoikas: It might be a square or some other odd type path :wink:

DoobesURU: > 
> 
> It might be a square or some other odd type path 😉

Good point! We artists tend to like to do crazy stuff.

DoobesURU:
Made some changes and they appear to work! You can still have two (or more) keyframes in the same location, but if the curve has no length, it stops.

korman/exporter/camera.py
Hoikas: Whoops, let's try:
```python
if any((len(i.keys) == 1 for i in (pos_ctrl.X, pos_ctrl.Y, pos_ctrl.Z) if i is not None)):
    raise ExportError("'{}': Rail Camera must have more than one keyframe", bo.name)
```

DoobesURU: Seems to work! I tried exporting with three keyframes (passed), one keyframe for all curves (stopped) and three keyframes for only the y curve (passed)

Issue #69, opened 2017-10-01
Automatic stencil at export from VCol layer


dpogue:
Setting up stencil layers with the blend ramp by hand kinda sucks, and you need to make sure to get the flags right.

PlasmaMax has a feature to allow artists to use vertex colour layers to define the blending stencils. Somewhere back in the pyprp days, either Nadnerb or myself (or both?) wrote a script to turn a VCol layer into a UV mapping for a stencil, but it wasn't automatic and could easily get out of sync.

My proposal is that we add a field to the per-layer Plasma panel with a checkbox for "Use vertex colour blending mask" and a select box for which VCol layer to apply as a stencil. At export time, this would automatically create the blending layer with the right texture, right flags, and UV coordinates.
For simplicity, we'd handle the VCol colours the way we do for alpha: a range from white to black.

Deledrius:
That sounds immensely useful and utilizes existing features well.

DoobesURU:
Just to add my two cents (over two years later...better late than never!), I'd very much like to see this. A lot of fan Ages, and now, converted Cyan Ages, use an Alpha VCol layer. Currently, one has to manually convert an alpha VCol to a texture, then set it up as a stencil, for things to work. Having an option that does it automatically, like in PyPRP, would be very helpful!

Deledrius:
This allows you to paint a stencil blend in Blender onto the vertex colors and automatically create a stencil image?  Not just the generated blendramps?

DoobesURU:
Not sure how it works, but in PyPRP, using VCol layers named "Col" and "Alpha", with the Alpha VCol using greyscale, it could produce a stencil. It's detailed a bit here: https://www.guildofwriters.org/wiki/Planting_Some_Grass#Vertex_painting

dpogue:
What @DoobesURU is describing (vertex alpha) *should* already be possible with korman, using the [same naming convention](https://github.com/H-uru/korman/blob/master/korman/exporter/mesh.py#L276) that PyPRP used. This has nothing to do with stenciling, per se.

---

This issue is unrelated to vertex alpha, and more related to generating UV coords against a blend ramp based on a vertex paint layer.

i.e., given an object having a material with 2 layers, rather than adding a blend ramp and manually setting up UV coordinates for it, you would instead add a vertex paint layer with greyscale values that would automatically be converted into UV values on a blend ramp at export time.

If the "stencil" vertex paint value of a vertex is black, generate a UV coordinate of 0,0.
[...]

PR #202, opened 2020-09-17
Ensure alpha vcol paint makes a BlendSpan

dpogue:
This seems to be the same flag Cyan [uses](https://github.com/H-uru/Plasma/blob/47f0899d1edc830ad54697f970b19639e0e8f098/Sources/Tools/MaxConvert/plMeshConverter.cpp#L1205-L1212) for marking geospans as needing to be in BlendSpans, conveniently enough for us

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -39,7 +39,7 @@ class _RenderLevel:
     _MAJOR_SHIFT = 28
     _MINOR_MASK = ((1 << _MAJOR_SHIFT) - 1)
 
-    def __init__(self, bo, hsgmat, pass_index, blend_span=False):
+    def __init__(self, bo, pass_index, blend_span=False):
         self.level = 0
         if pass_index > 0:
             self.major = self.MAJOR_FRAMEBUF
@@ -71,8 +71,8 @@ class _RenderLevel:
 
 
 class _DrawableCriteria:
-    def __init__(self, bo, hsgmat, pass_index):
-        self.blend_span = bool(hsgmat.layers[0].object.state.blendFlags & hsGMatState.kBlendMask)
+    def __init__(self, bo, geospan, pass_index):
+        self.blend_span = bool(geospan.props & plGeometrySpan.kRequiresBlending)
         self.criteria = 0
 
         if self.blend_span:
@@ -80,7 +80,7 @@ class _DrawableCriteria:
                 self.criteria |= plDrawable.kCritSortFaces
             if self._span_sort_allowed(bo):
                 self.criteria |= plDrawable.kCritSortSpans
-        self.render_level = _RenderLevel(bo, hsgmat, pass_index, self.blend_span)
+        self.render_level = _RenderLevel(bo, pass_index, self.blend_span)
 
     def __eq__(self, other):
         if not isinstance(other, _DrawableCriteria):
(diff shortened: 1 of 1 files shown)


korman/exporter/mesh.py
```diff
@@ -218,6 +218,11 @@ def _create_geospan(self, bo, mesh, bm, hsgmatKey):
         geospan = plGeometrySpan()
         geospan.material = hsgmatKey
 
+        # Mark us as need a BlendSpan if the material require blending
```
Deledrius, line 221: Grammar (needing). :)

dpogue: Something terribly has gone wrong

Deledrius (approved):

DoobesURU:
I've made the changes to the script and have tried testing it. Is there a way to tie the alpha vcol value to the AlphaBlend texture(s) that Cyan uses? I'm guessing that is their way of saying "the vcol alpha goes between/after these textures". Am I missing something?

dpogue:
@DoobesURU No, that's not how it works. Blending between textures within an object will always need to use stencils. (#69 is about automatically generating a stencil from a vertex paint layer, but that's a totally separate idea.)

Alpha vertex paint works just like normal vertex paint, except that you can paint a vertex to be partially or fully transparent.

korman/exporter/mesh.py
```diff
@@ -218,6 +218,11 @@ def _create_geospan(self, bo, mesh, bm, hsgmatKey):
         geospan = plGeometrySpan()
         geospan.material = hsgmatKey
 
+        # Mark us as needing a BlendSpan if the material require blending
+        blend_span = bool(hsgmatKey.object.layers[0].object.state.blendFlags & hsGMatState.kBlendMask)
+        if blend_span:
```
Hoikas, line 223: Might be best to just combine this all into one line and drop the explicit conversion to a bool.

korman/exporter/mesh.py
```diff
@@ -430,6 +435,10 @@ def _export_geometry(self, bo, mesh, materials, geospans):
                     uvMap[numUVs - 1].normalize()
                     vtx.uvs = uvMap
 
+            # Mark us for blending if we have a alpha vertex paint layer
+            if alpha is not None:
+                geospan.props |= plGeometrySpan.kRequiresBlending
```
Hoikas, line 440: This seems a little overzealous. What if an artist is in the habit of adding white alpha vcol layers to all their meshes?

Hoikas (commented):
This looks like the right approach... Just a few notes.

korman/exporter/mesh.py
DoobesURU: > What if an artist is in the habit of adding white alpha vcol layers to all their meshes?

You mean like Cyan? 😉

Jrius:
> @DoobesURU No, that's not how it works. Blending between textures within an object will always need to use stencils. (#69 is about automatically generating a stencil from a vertex paint layer, but that's a totally separate idea.)
> 
> Alpha vertex paint works just like normal vertex paint, except that you can paint a vertex to be partially or fully transparent.

I figured I would jump into the conversation: it seems using vertex alpha to blend between two textures of the same object/material DOES happen in Kemo. For instance, on object `Garden2`. Vertices have alpha painting, which is used instead of a stencil between layers 2990 and 301/302 of material `Material #79_3`. Haven't checked in detail how this works, but I suspect this has to do with layer 2990 having no alpha (hence being drawn opaque), and 301/302 having alpha blending (thus being affected by vertex paint alpha when draw over the first layer).

korman/exporter/mesh.py
Jrius: > You mean like Cyan? wink

Not the fault of Cyan's artists. This is actually an issue with ZLZ. For various reasons, the PRP format just expects all drawable vertices to have all four RGBA color channels, and ZLZ reflects that when importing Cyan's Ages. So ZLZ will always create two vertex color layers, even when all vertices are fully white or opaque and thus the color/alpha layer is not needed.

I could just tell ZLZ to skip creating those layers when all vertices are white/opaque. This has never bothered me before, but I admit it would make imported Ages much cleaner.

dpogue:
Updated to check if any of the "Alpha" vertex colours are actually < 1.0 before flagging for blending.

Hoikas (approved):

PR #207, opened 2020-09-25
Add Runtime Light Option

DoobesURU:
Adds kLPMovable to lights without animation for stationary real-time lamps.


DoobesURU:
Closing this as kLPMovable should be a bit more "special". ;)

PR #208, opened 2020-09-25
Add Runtime Color Setting

DoobesURU:
Adds a FloatVectorProperty into materials to set the Runtime color. Also sets it to export properly. Adds new properties and ui scripts for materials window.


DoobesURU:
I'd like to be able to wedge this either between or just under Diffuse and Specular, but I'm not sure how I'd do that. 😉

DoobesURU:
Closing as Hoikas is working on another means to regulate runtime in #210 .

PR #209, opened 2020-09-27
Add Double Sided Toolbox Operators

DoobesURU:
Adds toolbar options to toggle meshes double sided, one to Disable All (handy for incoming PyPRP Ages) and Enable/Disable Selection.

I thought about an Enable All button, but didn't know if there would ever be a reason to use a function like that.

### DIFF
--- a/korman/operators/op_toolbox.py
+++ b/korman/operators/op_toolbox.py
@@ -171,6 +171,37 @@ class PlasmaToggleAllPlasmaObjectsOperator(ToolboxOperator, bpy.types.Operator):
             i.plasma_object.enabled = self.enable
         return {"FINISHED"}
 
+        
+class PlasmaToggleDoubleSidedOperator(ToolboxOperator, bpy.types.Operator):
+    bl_idname = "mesh.plasma_toggle_double_sided"
+    bl_label = "Toggle All Double Sided"
+    bl_description = "Toggles all meshes to be double sided"
+    
+    enable = BoolProperty(name="Enable", description="Enable Double Sided")
+    
+    def execute(self, context):
+        enable = self.enable
+        for mesh in bpy.data.meshes:
+            mesh.show_double_sided = enable
+        return {"FINISHED"}
+
+
+class PlasmaToggleDoubleSidedSelectOperator(ToolboxOperator, bpy.types.Operator):
+    bl_idname = "mesh.plasma_toggle_double_sided_selected"
+    bl_label = "Toggle Selected Double Sided"
+    bl_description = "Toggles selected meshes double sided value"
+    
+    @classmethod
+    def poll(cls, context):
+        return super().poll(context) and hasattr(bpy.context, "selected_objects")
+
+    def execute(self, context):
+        mesh_list = [i.data for i in context.selected_objects if i.type == "MESH"]
+        enable = not all((mesh.show_double_sided for mesh in mesh_list))
+        for mesh in mesh_list:
+            mesh.show_double_sided = enable
+        return {"FINISHED"}
+
 
 class PlasmaToggleEnvironmentMapsOperator(ToolboxOperator, bpy.types.Operator):
     bl_idname = "texture.plasma_toggle_environment_maps"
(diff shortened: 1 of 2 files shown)


korman/operators/op_toolbox.py
```diff
@@ -170,7 +170,40 @@ def execute(self, context):
...
+    bl_idname = "mesh.plasma_toggle_double_sided"
+    bl_label = "Toggle All Double Sided"
+    bl_description = "Toggles all meshes to be double sided"
+    
+    enable = BoolProperty(name="Enable", description="Enable Double Sided")
+    
+    def execute(self, context):
+        enable = self.enable
+        for mesh in bpy.data.meshes:
+            if mesh is None:
```
Hoikas, line 184: I don't think this will ever happen. (Gulp)

korman/operators/op_toolbox.py
```diff
@@ -170,7 +170,40 @@ def execute(self, context):
...
+                continue
+            for material in mesh.materials:
+                if material is None:
+                    continue
```
Hoikas, line 188: I don't think this will actually filter out regions properly. It might be best to just remove it so the operator "does what it says"(TM).

korman/operators/op_toolbox.py
```diff
@@ -170,7 +170,40 @@ def execute(self, context):
...
+    
+    @classmethod
+    def poll(cls, context):
+        return super().poll(context) and hasattr(bpy.context, "selected_objects")
+
+    def execute(self, context):
+        mesh_list = [i.data for i in context.selected_objects if i.type == "MESH"]
+        enable = not all((mesh.show_double_sided for mesh in mesh_list))
+        for mesh in mesh_list:
+                mesh.show_double_sided = enable
```
Hoikas, line 205: Too much spacing :wink:

korman/ui/ui_toolbox.py
```diff
@@ -50,6 +50,13 @@ def draw(self, context):
         col.operator("texture.plasma_toggle_environment_maps", icon="IMAGE_RGB", text="Enable All EnvMaps").enable = True
         col.operator("texture.plasma_toggle_environment_maps", icon="IMAGE_RGB_ALPHA", text="Disable All EnvMaps").enable = False
 
+        # Double Sided Operators
+        col.label("Double Sided:")
+        col.operator("mesh.plasma_toggle_double_sided", icon="MESH_DATA", text="Disable All").enable = False
+        mesh_list = [i.data for i in bpy.context.selected_objects if i.type == "MESH"]
+        all_double_sided = all((mesh.show_double_sided for mesh in mesh_list))
```
Hoikas, line 57: You could simplify this by merging these two lines:
```python
all_double_sided = all((mesh.show_double_sided for i in bpy.context.selected_objects if i.type == "MESH"))
```

DoobesURU: Shouldn't that be `i.show_double_sided`?

Hoikas: Oh, yes. Whoops.

DoobesURU: I see where you were going with that and devised this:
`all_double_sided = all((mesh.show_double_sided for mesh in [i.data for i in bpy.context.selected_objects if i.type == "MESH"]))`

Hoikas: I was hoping to get rid of the list part `[i.data for i in bpy.context.selected_objects if i.type == "MESH"]` because it stores all the meshes then processes them. I see what I have wrong in the code I gave you previously...

```python
all_double_sided = all((i.data.show_double_sided for i in bpy.context.selected_objects if i.type == "MESH"))
```

korman/operators/op_toolbox.py
```diff
@@ -170,7 +170,35 @@ def execute(self, context):
...
+    @classmethod
+    def poll(cls, context):
+        return super().poll(context) and hasattr(bpy.context, "selected_objects")
 
+    def execute(self, context):
+        mesh_list = [i.data for i in context.selected_objects if i.type == "MESH"]
+        enable = not all((mesh.show_double_sided for mesh in mesh_list))
+        for mesh in mesh_list:
+            mesh.show_double_sided = enable
+        return {"FINISHED"}
```
Hoikas, line 202: OK, last nightpick :tongue: [PEP8](https://www.python.org/dev/peps/pep-0008/) says that between each `class`, there should be two blank lines.

DoobesURU: Done!

Hoikas (approved):

PR #210, opened 2020-09-29
DSpan Export Improvements

Hoikas:
This is designed to fix some issues around non-animated runtime lights not affecting geometry and challenges around alpha transparency and x-ray effects. We do so by properly using the `kLiteVtxNonPreshaded` equation for alpha and runtime lit geometry. Further, preshade and runtime colors are now properly set for runtime lit and emissive layers.

Adds the Render > Blending modifier to ensure dependent objects get sorted into later passes.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -147,16 +147,6 @@ class MaterialConverter:
             "transformCtl": self._export_layer_transform_animation,
         }
 
-    def calc_material_ambient(self, bo, bm) -> hsColorRGBA:
-        emit_scale = bm.emit * 0.5
-        if emit_scale > 0.0:
-            return hsColorRGBA(bm.diffuse_color.r * emit_scale,
-                               bm.diffuse_color.g * emit_scale,
-                               bm.diffuse_color.b * emit_scale,
-                               1.0)
-        else:
-            return utils.color(bpy.context.scene.world.ambient_color)
-
     def _can_export_texslot(self, slot):
         if slot is None or not slot.use:
             return False
@@ -1176,15 +1166,25 @@ class MaterialConverter:
     def get_bump_layer(self, bo):
         return self._bump_mats.get(bo, None)
 
+    def get_material_ambient(self, bo, bm) -> hsColorRGBA:
+        emit_scale = bm.emit * 0.5
+        if emit_scale > 0.0:
+            return hsColorRGBA(bm.diffuse_color.r * emit_scale,
+                               bm.diffuse_color.g * emit_scale,
+                               bm.diffuse_color.b * emit_scale,
+                               1.0)
+        else:
+            return utils.color(bpy.context.scene.world.ambient_color)
+
     def get_material_preshade(self, bo, bm, color=None) -> hsColorRGBA:
-        if bo.plasma_modifiers.lighting.rt_lights or bm.emit:
+        if bo.plasma_modifiers.lighting.rt_lights:
             return hsColorRGBA.kBlack
         if color is None:
             color = bm.diffuse_color
         return utils.color(color)
 
     def get_material_runtime(self, bo, bm, color=None) -> hsColorRGBA:
-        if bm.emit:
+        if not bo.plasma_modifiers.lighting.rt_lights:
             return hsColorRGBA.kBlack
         if color is None:
             color = bm.diffuse_color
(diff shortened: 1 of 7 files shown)


korman/properties/modifiers/base.py
```diff
@@ -30,10 +30,30 @@ def created(self):
...
         pass
 
+    @property
+    def draw_opaque(self):
+        """Render geometry before the avatar"""
+        return False
+
+    @property
+    def draw_framebuf(self):
+        """Render geometry after the avatar but before regular geometry"""
```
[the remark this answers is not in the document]
Hoikas: The avatar is drawn at the end of the opaque pass, immediately before the framebuf pass. The framebuf pass generally begins blended content, from what I can tell... although there are exceptions :/

DoobesURU:
I was able to get the Pub exported with the latest commit. It looks like things exported fine for the most part except the lamp flares.

It looks like flares that are _not_ marked "Shadeless" and have Emit > 0...don't emit and are, in fact, partially transparent, I think? This is without any vertex coloring, Col or Alpha.

Note that marking them shadeless fixes this somewhat, but they will not be able to go any brighter (and more realistic) without the emit property.

Hoikas:
One thing I noticed is that blender's default diffuse color has its intensity... or something set to non-full. I'm not sure what it's called, but it's the rightmost slider on the color picker.

DoobesURU:
> 
> 
> One thing I noticed is that blender's default diffuse color has its intensity... or something set to non-full. I'm not sure what it's called, but it's the rightmost slider on the color picker.

Yeah, that's the intensity. The flares in question did have that set to < 1, but I tried exporting one of the materials with an Intensity of 1 and emit of .75. Still came out with about half brightness (I was mistaken about transparency before).

Same settings with the Kirel flares and same result. /shrug

Hoikas:
Can you share a comparison screenshot so I can better understand what's happening? What might be happening is that a change I made to better match the Max plugin - setting the preshade color to black on emissive layers - might be biting us.

DoobesURU:
> 
> 
> Can you share a comparison screenshot so I can better understand what's happening? What might be happening is that a change I made to better match the Max plugin - setting the preshade color to black on emissive layers - might be biting us.

For posterity:
https://www.dropbox.com/s/rd0w0n6rznxiq7v/PubKorman210_01.png?dl=0
https://www.dropbox.com/s/w0fewsx5bjozl2z/PubKorman210_02.png?dl=0

DoobesURU:
Another issue I noticed: graphics get a bit twitchy when footprint decals appear after stepping out of water onto any object with alpha values that receives footprints.

korman/exporter/material.py
```diff
@@ -147,6 +147,16 @@ def __init__(self, exporter):
             "transformCtl": self._export_layer_transform_animation,
         }
 
+    def calc_material_ambient(self, bo, bm) -> hsColorRGBA:
+        emit_scale = bm.emit * 0.5
```
DoobesURU, line 151: OK, not sure how this is working (the color is odd in PRPShop), but this appears to correct the colors on my end. Maybe give it a whirl and see how it does?:

```
        emit_scale = (bm.emit / 10)
        if emit_scale > 0.0:
            return hsColorRGBA(bm.diffuse_color.r * (emit_scale + 1.0) + 0.1,
                               bm.diffuse_color.g * (emit_scale + 1.0) + 0.1,
                               bm.diffuse_color.b * (emit_scale + 1.0) + 0.1,
                               1.0)
```

DoobesURU: Per Hoikas' suggestion on IRC of tying shadeless into this equation, I tried this:

```
        if bm.emit > 0.0 or bm.use_shadeless == True:
            return hsColorRGBA(bm.diffuse_color.r + 0.175,
                               bm.diffuse_color.g + 0.175,
                               bm.diffuse_color.b + 0.175,
                               1.0)
```
Seems to work about the same if we're using Cyan's Ages.

dpogue: That seems like it could cause issues for brighter colours, potentially pushing them above 1.0?

I'm confused why we can't just use the diffuse colour?

Hoikas: I was hoping that avoiding the multiplication (and comparison) against `emit` would fix the colors. We still may have to assign the diffuse value to either `runtime` or `preshade`... Even though PlasmaMax does not...

DoobesURU: Using the diffuse color straight up causes the flares to be a slightly darker shade from their original counterparts...in this case, Kirel. It's really noticeable with the blue light garden lamp flares. The math above got them a lot closer FWIW.

Hoikas:
Per feedback in IRC, this is now functioning correctly.

PR #212, opened 2020-10-01
Add random sound option to Sound Emitter modifier.

Deledrius:


Deledrius:
That was my initial intent, but I had difficulty seeing how that would work (and was unable to find any examples in the game data), so I decided to make it an option on existing sounds.  Kickables use the random sound groups?

Hoikas:
Yes, see Personal_District_psnlMYSTII and GlobalAvatars_District_Audio - look at the random sounds for footstep sounds and plPhysicalSndGroup for examples of how this works on the engine level.

Deledrius:
Excellent!  Thanks, I'll check those.

DoobesURU:
Just to revive this discussion, I've done a bit of tinkering using @Deledrius ' code and created a separate Random Sound Mod. It's very basic right now (ie no sanity checks), but it's there.

Before I test it, I'm curious what the "Mode" setting is for in the parameters. I see Del has the Min and Max values adjustable, but nothing else. Is the mode relevant to something, and if so, what? Also, is there anything else that would need user adjustment?

PR #213, opened 2020-10-03
Update render.py

TikiBear:
Add flag to expose missing pivots to swivel interface. Requires update to the Modifiers render.py.


TikiBear:
When in the new GOME pub I muse that I would make the nimbus on each table lamp (candle?) rotate very slowly. This swivel fix is required to do that. My point is just to mention something that can't be done without this fix.

Deledrius:
Thanks for the reminder.  I'll try to review this soon. :)  Sorry for the wait.

Hoikas:
Merged in #214, closing.

PR #214, opened 2020-10-03
ViewFaceMod Changes

TikiBear:
Add offsets and pivots to the swivel interface. Requires update to UI render.py.

### DIFF
--- a/korman/properties/modifiers/render.py
+++ b/korman/properties/modifiers/render.py
@@ -513,6 +513,10 @@ class PlasmaViewFaceMod(idprops.IDPropObjectMixin, PlasmaModifierProperties):
                               description="Swivel only around the local Y axis",
                               default=False)
 
+    # TikiBear - add the two missing pivots
+    special_pivots = BoolProperty(name="Special Pivots",
+                              description="Default: 3-axis; With Pivot on local Y: 2-axis, local Y azimuth axis",
+                              default=False)
     offset = BoolProperty(name="Offset", description="Use offset vector", default=False)
     offset_local = BoolProperty(name="Local", description="Use local coordinates", default=False)
     offset_coord = FloatVectorProperty(name="", subtype="XYZ")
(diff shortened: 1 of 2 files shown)


Hoikas:
Thanks for the pull request, @TikiBear! I am asking @Deledrius to take a look at these changes because he was the original author of the swivel modifier, and I don't feel comfortable enough to make any judgement calls around its behavior. I just pick the pythonic nits in those cases :smile:

TikiBear:
Thank you! I look forward to you guiding me through Pythonic nits. I carry the stain of C++!

Here is the docent info for your visit to the Sprite Gallery.

The Sprite Gallery in Deep Island is actually designed to demonstrate the changes I propose. Each lamp has a swivel nimbus.The nimbus has rays which make it orientable. By observing the orientation as you move around you can see how manyaxes the different pivots have. The 3-axis pivot lacks the singularities plaguing the 2-axis pivots.

The imagers in The Gallery use local offsets to make them face the screen not the camera. The sconces use globaloffsets to prevent the nimbus from intersecting the supporting wall.

So far, no visitor to The Gallery has commented on these effects.
 
-----Original Message-----

Deledrius:
I'm not quite understanding what this fixes.  Can you explain (perhaps with one or more illustrations)?  Adding this option seems reasonable, but I'd like to know what the use-case is that you are seeing it applies to.

TikiBear:
In the Sprite Gallery in DI or UAM:

I use the local offset feature to have imagers face the screen rather than the camera. This also enables the avatar to walk through the image without encountering an axis-crossing singularity. You can't really demonstrate the absence of singularities in still pictures, though.  There are three imagers visible in [Imagers.jpg].

I use the global offset feature to keep each sconce's nimbus from intersecting the wall in the disco. [SconceFar.jpg, SconceClose.jpg]

I use the 3-axis pivot for the main lamps. The 3-axis pivot has no axis-crossing singularity. All other pivots have them (though note above scenario). [MainLamp.jpg]

The disco ball nimbus uses the 3-axis pivot for the above plus it rotates along with the disco ball. If you look up at the ball it rotates clockwise and so does the nimbus. If you look down the ball rotates counter-clockwise and so does the nimbus. Don't ask me to explain the geometry. I haven't figured it out, yet. [DiscoBall.jpg]

I acknowledge that I can't see a difference between the 2-axis on Z and 2-axis on Y pivots in-game. In the Blender (or other tool) files the nimbus must face along the azimuthal axis of 2-axis on Z pivots. For the 2-axis on Y pivots that's not the case. I would prefer to see them "looking up" in the Blender file. So, my guess is that the 2-axis on Y pivot is more for design than deployment.

Link to image folder: https://app.box.com/s/ocurcdqbe1oxxf5db8m09mzx1od0az5e

Deledrius:
Thanks for the additional explanation.  I tried to model the existing system closely on the presentation available in the 3DS Max plugin, and then streamlined it from there to highlight the typical use cases.  This particular modifier is a bit of a strange one to wrap the brain around behind-the-scenes.

TikiBear:
It isn't very intuitive. It was only after making and seeing the various scenarios that I understood the possibilities the features create.

Hoikas:
Trying to clear out old PRs. After 6 years more UX experience under my belt, this very much feels like a magic button style personal modification. There's nothing wrong with folks adding those for their own use, but it's not very discoverable or self documenting. That's against the spirit of the first bullet point of the Zen of Korman and is the same reason #258 hasn't been accepted.

With all that said, I think I'm going to close this one out. If anyone who understands the feature that's being added here wants to come up with a way to expose this in a discoverable way, potentially be redesigning the swivel modifier (and seamlessly upgrading old modifiers), that would be welcome.

TikiBear:
This leaves me as the only one who can make real fire marbles, since they have a special swivel. I'm fine with that.

PR #215, opened 2020-10-04
Fix bugs caused by using Decimate modifier

TikiBear:
Prevent export of Blender Decimate modifier from failing and corrupting Blender file!

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -174,11 +174,12 @@ class _MeshManager:
             trash_mesh, bo.data = bo.data, data_meshes.get(override["mesh"])
             data_meshes.remove(trash_mesh)
 
-            # If modifiers were removed, reapply them now.
+            # If modifiers were removed, reapply them now unless they're read-only.
+            readonly_attributes = {("DECIMATE", "face_count"),}
             for cached_mod in override["modifiers"]:
                 mod = bo.modifiers.new(cached_mod["name"], cached_mod["type"])
                 for key, value in cached_mod.items():
-                    if key in {"name", "type"} or (cached_mod["type"] == "DECIMATE" and key=="face_count"): # Decimate attribute face_count is read-only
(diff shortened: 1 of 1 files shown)


korman/exporter/mesh.py
```diff
@@ -178,7 +178,7 @@ def __exit__(self, type, value, traceback):
             for cached_mod in override["modifiers"]:
                 mod = bo.modifiers.new(cached_mod["name"], cached_mod["type"])
                 for key, value in cached_mod.items():
-                    if key in {"name", "type"}:
+                    if key in {"name", "type"} or (cached_mod["type"] == "DECIMATE" and key=="face_count"): # Decimate attribute face_count is read-only
```
Hoikas, line 181: I saw this recently when making a sample Age for ametist and was wondering what was going on. I think we can make this fix a little more flexible by making a tuple of `(cached_mod["type"], key)` and comparing that against a sequence of known read-only attributes... In other words:

```python
readonly_attribs = {("DECIMATE", "face_count"),}
# ... (some code here) ...
if key in {"name", "type"} or (cached_mod["type"], key) in readonly_attribs:
    continue
```

That way if we find any more read-only attributes like `face_count`, it's fairly trivial to add them.

Hoikas (commented):
Thanks for the pull request! It looks like you're using the github web UI to submit patches. In my experience, this is a little bit challenging to work with. You may want to investigate a git GUI such as [gitkraken](https://www.gitkraken.com/) or [fork](https://git-fork.com/) so you can work directly against the files on your computer.

korman/exporter/mesh.py
TikiBear: I like the tupling idea. I'll make an update. I'll go get a git GUI.

TikiBear: Getting a git GUI turns out to be a problem for my old PC. Oh, well.

TikiBear: I have updated per suggestion.

Hoikas (approved):
LGTM

PR #216, opened 2020-10-04
Correct layer animation timing for FPS other than 30

TikiBear:
Correct layer animation timing for FPS other than 30.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -535,8 +535,13 @@ class MaterialConverter:
             start, end = functools.reduce(lambda x, y: (min(x[0], y[0]), max(x[1], y[1])),
                                           (fcurve.range() for fcurve in fcurves))
 
-            atc.begin = start / fps
-            atc.end = end / fps
+            # TikiBear - propagate adjustment from AnimationConverter._process_fcurve
+            if fps == 30.0:
+                atc.begin = start / fps
+                atc.end = end / fps
+            else:
+                atc.begin = (start * 30) / (fps * fps)
+                atc.end = (end * 30) / (fps * fps)
 
             layer_props = tex_slot.texture.plasma_layer
             if not layer_props.anim_auto_start:


Hoikas (commented):
I'm a little confused on this pull request. From what I can tell, `atc.begin` and `atc.end` are frame times, and `bpy.types.FCurve.range()` should also return frame times. So, I'm not sure why we were originally dividing by the fps. Wouldn't just using the start and end values from `range()` work? Our animation converter code is doing unspeakable things, but it shouldn't be changing the FCurves themselves. :/

TikiBear:
Texture animation empirically doesn't work properly without this change, at least for me. The animation ends too early or too late at other frame rates. My feeling is this just compensates for something unspeakable done earlier!

TikiBear:
It's possible that I'm looking at the timing incorrectly. Let's hold off on this for now. It's not a big priority anyway.

TikiBear:
This is what I see without the fix: Sprite Gallery has both physical and texture animations. Normally I have FPS=30. If I change to FPS=60 the physical animations run twice as fast and last half the normal duration. The texture animations run twice as fast, but have the same duration, so for half of the duration there's no animation. With the fix the duration is also shortened and there's no dead space. So, for me the fix corrects a problem.

Deledrius:
Thanks, that's a very specific example we can look into!

TikiBear:
The texture animation is a loop, so you can easily see that half of the loop is static.

Hoikas:
This fix is correct - Blender's documentation is misleading about `FCurve.range()`. However, the math submitted here is a bit hard to follow, so I opted reformat it to more closely match the animation converter (the fps squared was confusing IMO). Also removed the use of `functools.reduce`.

TikiBear:
Indeed, it was the appearance of fps squared in the original that made me suspicious of the math in the first place. I'm glad that a more-legible formula could be found!

PR #217, opened 2020-10-04
Update convert.py

TikiBear:
Fix overwritten "layers = pages paradigm" default pages.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -162,6 +162,18 @@ class Exporter:
         error = explosions.UndefinedPageError()
         for obj in scene.objects:
             if obj.plasma_object.enabled:
+                if not obj.plasma_object.page:
+                    # this was initialized but overwritten
+                    # We want to encourage the pages = layers paradigm.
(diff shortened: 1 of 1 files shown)


dpogue:
I think I would argue that most people don't use Blender layers for different Plasma pages. If anything, it's more likely people would use Blender scenes (although those are slightly annoying to work with).

Speaking for myself, I use Blender layers to group different types of objects: geometry in one layer; colliders and regions in another; lamps in another; OneShotMods, sound emitters, and spawn points in another; etc.

Deledrius:
I agree with @dpogue.  That's how I use layers in Blender with korman as well.

TikiBear:
I don't know the originator of this. I wondered if it might be obsolete now. It's still present in the initialization code.
 
 
-----Original Message-----

Deledrius:
> I don't know the originator of this. I wondered if it might be obsolete now. It's still present in the initialization code.
@Hoikas can elaborate, but as I understand it there was an intent to support that model initially but it has not been the case for some time.

PR #218, opened 2020-10-07
Update subworld handling of collisions and animation

TikiBear:
Update subworld handling of collisions and animation.

### DIFF
--- a/korman/exporter/physics.py
+++ b/korman/exporter/physics.py
@@ -157,7 +157,7 @@ class PhysicsConverter:
 
             # Got subworlds?
             subworld = bo.plasma_object.subworld
-            if self.is_dedicated_subworld(subworld, sanity_check=False):
+            if subworld is not None and self.is_dedicated_subworld(subworld, sanity_check=False):
                 physical.subWorld = self._mgr.find_create_key(plHKSubWorld, bl=subworld)
 
             # Export the collision modifier here since we like stealing from it anyway.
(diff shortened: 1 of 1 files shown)


TikiBear:
This is what I did to make subworlds work in The Flying Cauldron and Duck Knee.

korman/exporter/physics.py
```diff
@@ -157,7 +157,7 @@ def generate_physical(self, bo, so, **kwargs):
 
             # Got subworlds?
             subworld = bo.plasma_object.subworld
-            if self.is_dedicated_subworld(subworld, sanity_check=False):
+            if subworld is not None and self.is_dedicated_subworld(subworld, sanity_check=False):
```
Hoikas, line 160: It looks like `is_dedicated_subworld()` already returns `False` for subworlds that are `None`.

korman/exporter/physics.py
```diff
@@ -212,11 +212,11 @@ def _iter_object_tree(bo, stop_at_subworld):
                 bo_xformed = bo.plasma_object.has_transform_animation
 
                 # MOUL: only objects that have animation data are kPhysAnim
-                if ver != pvMoul or bo_xformed:
+                if mod.enabled and (ver != pvMoul or bo_xformed):
```
Hoikas, line 215: This could break animated clickables and regions because their physics objects are exported by nodes, not the collision modifier.

korman/exporter/physics.py
```diff
@@ -212,11 +212,11 @@ def _iter_object_tree(bo, stop_at_subworld):
...
                     _set_phys_prop(plSimulationInterface.kPhysAnim, simIface, physical)
                 # PotS: objects inheriting parent animation only are not pinned
                 # MOUL: animated objects in subworlds are not pinned
-                if bo_xformed and (ver != pvMoul or subworld is None):
+                if (bo_xformed and (ver != pvMoul or subworld is None)) or ((ver != pvMoul) and subworld is not None and (not bo_xformed)):
```
Hoikas, line 219: I suspect this is what's fixing subworlds, but the `kPinned` property only has an effect in PotS. The conditional is really hard to follow, so I'm not really able to glance in and see what's happening.

Hoikas:
@julyfortoday Did some work on demonstrating the subworld problems on the forum. I wasn't really able to understand his sample blend though, and edits to his posts indicated that he had figured out a number of the problems. Maybe he'll chime in? *cross fingers*

TikiBear:
It sounds as though my approach is too naive. I'm fine standing down. I can only test on PotS, anyway. With everyone's helpful input I can watch for unexpected side effects going forward.

Issue #204, opened 2020-09-23
Day/night cycle animation support


dpogue:
Day/Night cycles (or other animations spanning the duration of the day) are handled using plAgeGlobalAnims. Instead of being time or frame based, these animations run from their starting frame to their ending frame over the course of the Age's day length (slight simplification, since it involves SDLs and time of day percentages).

The goal is to add a new animation modifier to korman to flag these sorts of animations. It should also have a field for specifying which SDL variable controls the animation. The bindings already exist in PyHSPlasma.

**Tasks:**

- [ ] Set up a new korman modifier for "Day cycle animations"
  - It should take the name of an SDL variable that controls the animation progress
- [ ] Set up export code to export the animation as a plAgeGlobalAnim

Hoikas:
This would probably make more sense as a single property on the animation modifier.

dpogue:
Checkbox on the existing modifier for "Driven by SDL variable" that enables a textbox for the SDL varname?

Hoikas:
I was thinking a single text field for the SDL variable--a blank input is a plain old ATC Anim.

DoobesURU:
The above PR goes along the lines of what Hoikas was thinking for the modifier: a single text field entry. At the moment, it's only enabled when the "Loop Anim" option is selected. Not sure if that's how we want it?

Issue #140, opened 2019-03-01
Multiple animations per object


Hoikas:
Presently, animations are rather inflexibly hardwired in for the Action currently linked to the object. In some cases, it is desirable to have multiple animations attached to a single object. As such, we should allow the user to specify Actions to export from the animation modifier or (other pertinent location). No setting should keep the old default behavior.

Hoikas:
Deferring to later release, potentially 0.11.

Hoikas:
After experimenting with NLA strips and looking over the PlasmaMax documentation, I have decided that Blender's NLA animation system does not meet the needs of Korman. Instead, we will retain the animation modifier and allow it to point to segments of the animation to export as completely separate animations, just like the Max plugin does. Multiple animations are mostly useful for GUIs, so the need to use them is probably limited, so there's no need for an over engineered solution.

Hoikas:
Closed by #282.

PR #219, opened 2020-10-09
Add plAgeGlobalAnim support

DoobesURU:
Adds SDL Global animation to Korman per #204 (work in progress).

What it does so far:

* Adds SDL field to Animation modifier. ~~Note that it's only available if "Loop Anim" is selected (is this OK?).~~
* Creates ~~what appears to be~~ an ~~valid~~ Age Global Animation (plAgeGlobalAnim) entry into the PRP files ~~aside from one flaw (see below)~~.

A few things to fix:

* ~~SDL parameter from modifier isn't being written to AgeGlobalAnim's "GlobalVarName" entry.~~ Fixed thanks to dpogue.
* ~~Global Animation being created for non-object animations (layer animations), causing glitching.~~
* ~~Some extra ATC parameters (says it needs fixing?)~~
* "Writing Age data..." part of export is hanging for several seconds, most likely due to mah hacks. 😉 
* ~~Global Animation also creates a near-empty ATCAnim.~~

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -72,7 +72,11 @@ class AnimationConverter:
         # There is a race condition in the client with animation loading. It expects for modifiers
         # to be listed on the SceneObject in a specific order. D'OH! So, always use these funcs.
         agmod, agmaster = self.get_anigraph_objects(bo, so)
-        atcanim = self._mgr.find_create_object(plATCAnim, so=so)
+        anim_data = bo.plasma_modifiers.animation
+        if anim_data.obj_sdl_anim:
+            atcanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
+        else:
+            atcanim = self._mgr.find_create_object(plATCAnim, so=so)
 
         # Add the animation data to the ATC
         for i in applicators:
@@ -89,21 +93,30 @@ class AnimationConverter:
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
-        atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
-        atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+        if anim_data.obj_sdl_anim:
+            atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, index=0)))
+            atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, index=1)))
+        else:
+            atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
+            atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
 
         # Marker points
-        if obj_action is not None:
+        if obj_action is not None and anim_data.obj_sdl_anim is None:
             for marker in obj_action.pose_markers:
                 atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
 
-        # Fixme? Not sure if we really need to expose this...
-        atcanim.easeInMin = 1.0
-        atcanim.easeInMax = 1.0
-        atcanim.easeInLength = 1.0
-        atcanim.easeOutMin = 1.0
-        atcanim.easeOutMax = 1.0
-        atcanim.easeOutLength = 1.0
+        def set_anim_params(plAGAnim):
+            sdl_name = bo.plasma_modifiers.animation.obj_sdl_anim
+            if sdl_name:
+                atcanim.globalVarName = sdl_name
+            else:
+                # Fixme? Not sure if we really need to expose this...
+                atcanim.easeInMin = 1.0
+                atcanim.easeInMax = 1.0
+                atcanim.easeInLength = 1.0
+                atcanim.easeOutMin = 1.0
+                atcanim.easeOutMax = 1.0
+                atcanim.easeOutLength = 1.0
 
     def _convert_camera_animation(self, bo, so, obj_fcurves, data_fcurves):
         if data_fcurves:
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -73,10 +73,7 @@ class AnimationConverter:
         # to be listed on the SceneObject in a specific order. D'OH! So, always use these funcs.
         agmod, agmaster = self.get_anigraph_objects(bo, so)
         anim_mod = bo.plasma_modifiers.animation
-        if anim_mod.obj_sdl_anim:
-            atcanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
-        else:
-            atcanim = self._mgr.find_create_object(plATCAnim, so=so)
+        atcanim = self._mgr.find_create_object(anim_mod.anim_type, so=so)
 
         # Add the animation data to the ATC
         for i in applicators:
@@ -93,26 +90,23 @@ class AnimationConverter:
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
-        def anim_range_type(plAGAnim):
-            if anim_mod.obj_sdl_anim:
-                atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
-                atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
-                atcanim.globalVarName = anim_mod.obj_sdl_anim
-            else:
-                atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
-                atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
-                # Fixme? Not sure if we really need to expose this...
-                atcanim.easeInMin = 1.0
-                atcanim.easeInMax = 1.0
-                atcanim.easeInLength = 1.0
-                atcanim.easeOutMin = 1.0
-                atcanim.easeOutMax = 1.0
-                atcanim.easeOutLength = 1.0
-
+        sdl_name = bo.plasma_modifiers.animation.obj_sdl_anim
+        atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
+        atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+        if isinstance(atcanim, plAgeGlobalAnim):
+            atcanim.globalVarName = anim_mod.obj_sdl_anim
+        if isinstance(atcanim, plATCAnim):
         # Marker points
-        if obj_action is not None and anim_mod is None:
-            for marker in obj_action.pose_markers:
-                atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
+            if obj_action is not None and not anim_data.obj_sdl_anim:
+                for marker in obj_action.pose_markers:
+                    atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
+            # Fixme? Not sure if we really need to expose this...
+            atcanim.easeInMin = 1.0
+            atcanim.easeInMax = 1.0
+            atcanim.easeInLength = 1.0
+            atcanim.easeOutMin = 1.0
+            atcanim.easeOutMax = 1.0
+            atcanim.easeOutLength = 1.0
 
     def _convert_camera_animation(self, bo, so, obj_fcurves, data_fcurves):
         if data_fcurves:
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -90,14 +90,14 @@ class AnimationConverter:
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
-        sdl_name = bo.plasma_modifiers.animation.obj_sdl_anim
+        sdl_name = anim_mod.obj_sdl_anim
         atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
         atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
         if isinstance(atcanim, plAgeGlobalAnim):
             atcanim.globalVarName = anim_mod.obj_sdl_anim
         if isinstance(atcanim, plATCAnim):
-        # Marker points
-            if obj_action is not None and not anim_data.obj_sdl_anim:
+            # Marker points
+            if obj_action is not None:
                 for marker in obj_action.pose_markers:
                     atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
             # Fixme? Not sure if we really need to expose this...
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -22,7 +22,7 @@ import weakref
 
 from . import utils
 
-class AnimationConverter:
+class AnimationConverter(plAGAnim):
     def __init__(self, exporter):
         self._exporter = weakref.ref(exporter)
         self._bl_fps = bpy.context.scene.render.fps
@@ -75,7 +75,7 @@ class AnimationConverter:
         anim_data = bo.plasma_modifiers.animation
         if anim_data.obj_sdl_anim:
             atcanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
-        else:
+        elif anim_data.obj_sdl_anim is None:
             atcanim = self._mgr.find_create_object(plATCAnim, so=so)
 
         # Add the animation data to the ATC
@@ -96,7 +96,7 @@ class AnimationConverter:
         if anim_data.obj_sdl_anim:
             atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, index=0)))
             atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, index=1)))
-        else:
+        elif anim_data.obj_sdl_anim is None:
             atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
             atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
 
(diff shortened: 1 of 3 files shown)


korman/exporter/animation.py
```diff
@@ -73,37 +73,55 @@ def fetch_animation_data(id_data):
...
 
         # Marker points
         if obj_action is not None:
             for marker in obj_action.pose_markers:
                 atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
 
+        def set_anim_params(plAGAnim):
+            sdl_name = bo.plasma_modifiers.animation.obj_sdl_anim
+            if sdl_name is not None:
+                globalanim.fGlobalVarName = sdl_name
```
dpogue, line 116: I think this should be `globalanim.globalVarName`

korman/exporter/animation.py
```diff
@@ -73,37 +73,55 @@ def fetch_animation_data(id_data):
         # to be listed on the SceneObject in a specific order. D'OH! So, always use these funcs.
         agmod, agmaster = self.get_anigraph_objects(bo, so)
         atcanim = self._mgr.find_create_object(plATCAnim, so=so)
+        globalanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
```
dpogue, line 76: Because a plAgeGlobalAnim is a special type of plATCAnim and has all the same functions available, you might be able to simplify this and avoid creating two objects and a bunch of these `if` conditions.

Something like this might work:
```python
if anim_data.obj_sdl_anim:
  atcanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
else:
  atcanim = self._mgr.find_create_object(plATCAnim, so=so)
```

Then you can mostly use `atcanim` without caring whether it's a plATCAnim or a plAgeGlobalAnim.

dpogue (commented):
Thanks for tackling this Doobes!

korman/exporter/animation.py
DoobesURU: Yep, that fix that part at least. Thank you!

DoobesURU: Actually, after redoing the script a bit, it's back to not working. ☹️

korman/exporter/animation.py
DoobesURU: I've reworked that section of the script a bit to try what you said and pushed the changes. It's now only creating a plAgeGlobalAnim for anything with an SDL state, but for some reason, it's still creating a supplementary ATCAnim object...likely due to the set_anim_params def I added?

dpogue: It's possible the plATCAnim is getting created by another modifier. There are 2 spots in properties/modifiers/anim.py that look like they create ATCAnims.

I can't see where you're actually calling the `set_anim_params` function...?

DoobesURU: I've taken out the `set_anim_params` def, but it's still creating an ATCAnim for the global animations. Is it possible it's giving the object a second pass and adding it then? Might explain the "write data files" lagging? Might also have to do with this part:

```
            # Fixme? Not sure if we really need to expose this...
            atcanim.easeInMin = 1.0
            atcanim.easeInMax = 1.0
            atcanim.easeInLength = 1.0
            atcanim.easeOutMin = 1.0
            atcanim.easeOutMax = 1.0
            atcanim.easeOutLength = 1.0
```

DoobesURU:
Lastest commit (c5552f8) fixes the duplicated entry in ATC Anim for global animations. While the object animation appears to be working properly now, it seems to have messed up a layer SDL animation on the same object.

For reference, here's the file with which I've been testing: https://www.dropbox.com/s/62d5p50kztugyi9/DayNightTest02.blend?dl=0

korman/exporter/animation.py
```diff
@@ -22,7 +22,7 @@
 
 from . import utils
 
-class AnimationConverter:
+class AnimationConverter(plAGAnim):
```
Hoikas, line 25: O.o

korman/exporter/animation.py
```diff
@@ -72,7 +72,11 @@ def fetch_animation_data(id_data):
         # There is a race condition in the client with animation loading. It expects for modifiers
         # to be listed on the SceneObject in a specific order. D'OH! So, always use these funcs.
         agmod, agmaster = self.get_anigraph_objects(bo, so)
-        atcanim = self._mgr.find_create_object(plATCAnim, so=so)
+        anim_data = bo.plasma_modifiers.animation
```
Hoikas, line 75: `anim_data` is playing with my mind a little bit. I think because there is a blender type named `AnimationData`. Could we change this to something like `anim_mod`?

korman/exporter/animation.py
```diff
@@ -89,21 +93,29 @@ def get_ranges(*args, **kwargs):
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
-        atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
-        atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+        if anim_data.obj_sdl_anim:
+            atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, index=0)))
+            atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, index=1)))
+        elif anim_data.obj_sdl_anim is None:
```
Hoikas, line 99: This will cause things like lamp color animations to not have the proper length in age global animations. While I do think `get_ranges` is a bit aggressive, I'm not sure this is the correct fix.

korman/properties/modifiers/anim.py
```diff
@@ -60,37 +60,44 @@ class PlasmaAnimationModifier(ActionModifier, PlasmaModifierProperties):
...
-                atcanim.initial = -1.0
-            if self.loop:
-                loop_start = markers.get(self.loop_start)
-                if loop_start is not None:
-                    atcanim.loopStart = _convert_frame_time(loop_start.frame)
+        anim_data = bo.plasma_modifiers.animation
+        if anim_data.obj_sdl_anim:
+            atcanim = exporter.mgr.find_create_object(plAgeGlobalAnim, so=so)
+        else:
+            atcanim = exporter.mgr.find_create_object(plATCAnim, so=so)
```
Hoikas, line 74: This pattern of checking `obj_sdl_anim` to figure out what kind of object to create is happening fairly often. We could hide this detail by  adding a property to the animation modifier, like so:
```python
@property
def anim_type(self):
    return plAgeGlobalAnim if self.obj_sdl_anim else plATCAnim
```
Then you could simplify some of the export code by doing things like:
```python
anim = mgr.find_create_object(anim_mod.anim_type, so=so)
if isinstance(anim, plAgeGlobalAnim):
    # ... code for age SDL animation here ...
else:
    # ... regular animation but not SDL animation here ...
# both here...
```

korman/exporter/animation.py
DoobesURU: Yeah, I kinda figured that'd catch your eye. Any suggestions on an alternative spot for that...or an alternative callback entirely?

Hoikas: I don't think you need this at all?

korman/exporter/animation.py
DoobesURU: Sure!

korman/exporter/animation.py
DoobesURU: Yeah, that was part of my hatchet job of trying to get ATC and global separate. I'll try adding the data_action back in.

korman/properties/modifiers/anim.py
DoobesURU: Done!

korman/exporter/animation.py
DoobesURU: plAGAnim was initially needed for the globalVarName. Otherwise, it couldn't find that property. I may have an alternative now thanks to one of your other suggestions. We'll see if it works.

korman/properties/modifiers/anim.py
```diff
@@ -60,37 +60,48 @@ class PlasmaAnimationModifier(ActionModifier, PlasmaModifierProperties):
...
-            if initial_marker is not None:
-                atcanim.initial = _convert_frame_time(initial_marker.frame)
-            else:
-                atcanim.initial = -1.0
-            if self.loop:
-                loop_start = markers.get(self.loop_start)
-                if loop_start is not None:
-                    atcanim.loopStart = _convert_frame_time(loop_start.frame)
+        anim_mod = bo.plasma_modifiers.animation
+        anim = exporter.mgr.find_create_object(anim_mod.anim_type, so=so)
```
Hoikas, line 74:
```suggestion
        atcanim = exporter.mgr.find_create_object(anim_mod.anim_type, so=so)
```

korman/properties/modifiers/anim.py
```diff
@@ -60,37 +60,48 @@ class PlasmaAnimationModifier(ActionModifier, PlasmaModifierProperties):
...
+        if isinstance(anim, plAgeGlobalAnim):
+            atcanim = exporter.mgr.find_create_object(plAgeGlobalAnim, so=so)
+        else:
+            atcanim = exporter.mgr.find_create_object(plATCAnim, so=so)
```
Hoikas, line 78: These are making duplicate animation objects. Instead, I would have just what's under `else:` as `if not isinstance(anim, plATCAnim):`

korman/ui/modifiers/anim.py
```diff
@@ -43,6 +43,9 @@ def animation(modifier, layout, context):
         col.enabled = modifier.loop
```
Hoikas, line 43:
```suggestion
        col.enabled = modifier.loop and modifier.obj_sdl_anim
```

korman/exporter/animation.py
```diff
@@ -72,7 +72,11 @@ def fetch_animation_data(id_data):
         # There is a race condition in the client with animation loading. It expects for modifiers
         # to be listed on the SceneObject in a specific order. D'OH! So, always use these funcs.
         agmod, agmaster = self.get_anigraph_objects(bo, so)
-        atcanim = self._mgr.find_create_object(plATCAnim, so=so)
+        anim_mod = bo.plasma_modifiers.animation
+        if anim_mod.obj_sdl_anim:
+            atcanim = self._mgr.find_create_object(plAgeGlobalAnim, so=so)
+        else:
+            atcanim = self._mgr.find_create_object(plATCAnim, so=so)
```
Hoikas, line 79: I would replace this with: 
```python
atcanim = self._mgr.find_create_object(anim_mod.anim_type, so=so)
```

korman/exporter/animation.py
```diff
@@ -89,22 +93,27 @@ def get_ranges(*args, **kwargs):
...
+                atcanim.easeInMin = 1.0
+                atcanim.easeInMax = 1.0
+                atcanim.easeInLength = 1.0
+                atcanim.easeOutMin = 1.0
+                atcanim.easeOutMax = 1.0
+                atcanim.easeOutLength = 1.0
 
         # Marker points
-        if obj_action is not None:
+        if obj_action is not None and anim_mod is None:
```
Hoikas, line 113:
```suggestion
        if obj_action is not None and not anim_data.obj_sdl_anim :
```

korman/exporter/animation.py
```diff
@@ -89,22 +93,27 @@ def get_ranges(*args, **kwargs):
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
-        atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
-        atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+        def anim_range_type(plAGAnim):
```
Hoikas, line 96: You've turned this into a function, and it's not being called anywhere, so I don't think this will export correctly. I think the start, end, and "FIXME" stuff should all go back to the way it was before this change request. Then you can add in right below `atcanim.end = ...`:

```python
if isinstance(atcanim, plAgeGlobalAnim):
    atcanim.globalVarName = anim_mod.obj_sdl_anim
```

DoobesURU:
More things changed, but now the globalVarName is not exporting again.

korman/exporter/animation.py
```diff
@@ -89,22 +93,27 @@ def get_ranges(*args, **kwargs):
...
+            if anim_mod.obj_sdl_anim:
+                atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
+                atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+                atcanim.globalVarName = anim_mod.obj_sdl_anim
+            else:
+                atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
+                atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
+                # Fixme? Not sure if we really need to expose this...
```
Hoikas, line 104: You could put this under an `if isinstance(atcanim, plATCAnim):`

Hoikas (commented):
OK, I actually looked at the difference between `plATCAnim` and `plAgeGlobalAnim` this time. Part of me wonders if we should have a drop-down for selecting between age-global and logic driven animations to hide the irrelevant pieces of the UI.

And, of course, there's the chaos of #140 looming at some point on the horizon... :cry:

korman/ui/modifiers/anim.py
DoobesURU: This caused everything below "Start Marker" to disappear if Loop Anim was selected. I'll play around with this some more. 😉 

EDIT: putting "and **not**" before modifier.obj_sdl_anim fixed it.

DoobesURU:
It's still slowing things up for a few seconds at the Writing Age data... part, but it's functioning as it should be now.

korman/properties/modifiers/anim.py
```diff
@@ -60,37 +60,46 @@ class PlasmaAnimationModifier(ActionModifier, PlasmaModifierProperties):
...
-            else:
-                atcanim.initial = -1.0
-            if self.loop:
-                loop_start = markers.get(self.loop_start)
-                if loop_start is not None:
-                    atcanim.loopStart = _convert_frame_time(loop_start.frame)
+        anim_mod = bo.plasma_modifiers.animation
+        atcanim = exporter.mgr.find_create_object(anim_mod.anim_type, so=so)
+        if not isinstance(atcanim, plAgeGlobalAnim):
+            atcanim = exporter.mgr.find_create_object(plATCAnim, so=so)
```
Hoikas, line 76:
```suggestion
```

We have already created the animation object, so no need to do it again.

korman/ui/modifiers/anim.py
```diff
@@ -32,6 +32,7 @@ def animation(modifier, layout, context):
         return
 
     split = layout.split()
+    col = layout.column()
```
Hoikas, line 35:
```suggestion
```

korman/exporter/animation.py
```diff
@@ -89,21 +90,23 @@ def get_ranges(*args, **kwargs):
...
-                atcanim.setMarker(marker.name, self._convert_frame_time(marker.frame))
-
-        # Fixme? Not sure if we really need to expose this...
-        atcanim.easeInMin = 1.0
-        atcanim.easeInMax = 1.0
-        atcanim.easeInLength = 1.0
-        atcanim.easeOutMin = 1.0
-        atcanim.easeOutMax = 1.0
-        atcanim.easeOutLength = 1.0
+            if obj_action is not None and not anim_data.obj_sdl_anim:
```
Hoikas, line 100:
```suggestion
            if obj_action is not None:
```

Since we know we're not an SDL animation, we can blow away that check.

korman/exporter/animation.py
```diff
@@ -89,21 +90,23 @@ def get_ranges(*args, **kwargs):
                 if i is not None:
                     yield i.frame_range[index]
         atcanim.name = "(Entire Animation)"
+        sdl_name = bo.plasma_modifiers.animation.obj_sdl_anim
```
Hoikas, line 93:
```suggestion
        sdl_name = anim_mod.obj_sdl_anim
```

korman/exporter/animation.py
```diff
@@ -89,21 +90,23 @@ def get_ranges(*args, **kwargs):
...
         atcanim.start = self._convert_frame_time(min(get_ranges(obj_action, data_action, index=0)))
         atcanim.end = self._convert_frame_time(max(get_ranges(obj_action, data_action, index=1)))
-
+        if isinstance(atcanim, plAgeGlobalAnim):
+            atcanim.globalVarName = anim_mod.obj_sdl_anim
+        if isinstance(atcanim, plATCAnim):
         # Marker points
```
Hoikas, line 99:
```suggestion
            # Marker points
```

Style nitpick: we want to make sure our indentation is correct :wink:

korman/ui/modifiers/anim.py
```diff
@@ -40,9 +41,13 @@ def animation(modifier, layout, context):
     if action:
         layout.prop_search(modifier, "initial_marker", action, "pose_markers", icon="PMARKER")
         col = layout.column()
-        col.enabled = modifier.loop
+        col.enabled = modifier.loop and not modifier.obj_sdl_anim
         col.prop_search(modifier, "loop_start", action, "pose_markers", icon="PMARKER")
         col.prop_search(modifier, "loop_end", action, "pose_markers", icon="PMARKER")
+        col = layout.column()
```
Hoikas, line 47: I'm wondering if we should just add in a `layout.separator()` here and make this the full width.

DoobesURU: This causes the field to vanish. 😐

Hoikas: Did you also change over the next two things from `col` to `layout`?

DoobesURU: Fixed now.

DoobesURU:
Update: Yep, there appear to be problems with animated objects that have "Shadeless" materials and no Bake Lighting modifier. As mentioned above, when exported, the lighting goes all out of whack and does not produce the needed shadeless material. Parenting the object to an empty removes the problem, but this is definitely a flaw.

PR #221, opened 2020-10-12
Special export features: Update material.py

TikiBear:
Texture offsets are mapped from Blender to Plasma at export. This is enabled in the export panel.


TikiBear:
Companion pull requests: animation.py, op_export.py, ui_world.py

Hoikas:
Superseded and closed by #243

PR #222, opened 2020-10-12
Special Export features: Update ui_world.py

TikiBear:
Add enabling check boxes for texture-transform-on-export and hobble-beziers-on-export features.
Companion pull requests: op_export.py, animation.py and material.py


TikiBear:
Propose to close IFO #243.

Issue #205, opened 2020-09-24
Responder Message Wait Issue


Hoikas:
Responder messages are inappropriately sensitive to the ordering of the internal connection list on the socket. For example, assume that Node A sends Messages B and C. Message B has a callback and miscellaneous descendant messages. If, in the internal links list of Node A, Message B occurs before Message C, then Message C blocks on completion of Message B, even though the UI indicates Message B is sent on completion of Node A.

PR #223, opened 2020-10-12
Special Export Features: Update animation.py

TikiBear:
Hobble broken beziers so they at least behave repeatably.

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -653,6 +653,7 @@ class AnimationConverter:
 
     def _process_fcurve(self, fcurve, convert=None):
         """Like _process_keyframes, but for one fcurve"""
+        age = bpy.context.scene.world.plasma_age
         keyframe_data = type("KeyFrameData", (), {})
         fps = self._bl_fps
         pi = math.pi
@@ -669,8 +670,13 @@ class AnimationConverter:
                 keyframe.frame_num = int(frame_num * (30.0 / fps))
             keyframe.frame_time = frame_num / fps
             if fkey.interpolation == "BEZIER":
-                keyframe.in_tan = -(value - fkey.handle_left[1])  / (frame_num - fkey.handle_left[0])  / fps / (2 * pi)
-                keyframe.out_tan = (value - fkey.handle_right[1]) / (frame_num - fkey.handle_right[0]) / fps / (2 * pi)
+                # beziers don't work properly so they are made to work consistently
+                if age.bad_beziers:
+                    keyframe.in_tan = 0.0
+                    keyframe.out_tan = 0.0
+                else:
+                    keyframe.in_tan = -(value - fkey.handle_left[1])  / (frame_num - fkey.handle_left[0])  / fps / (2 * pi)
+                    keyframe.out_tan = (value - fkey.handle_right[1]) / (frame_num - fkey.handle_right[0]) / fps / (2 * pi)
                 bezier = True
             else:
                 keyframe.in_tan = 0.0
@@ -795,6 +801,7 @@ class AnimationConverter:
 
     def _process_keyframes(self, fcurves, convert=None):
         """Groups all FCurves for the same frame together"""
+        age = bpy.context.scene.world.plasma_age
         keyframe_data = type("KeyFrameData", (), {})
         fps = self._bl_fps
         pi = math.pi
@@ -824,8 +831,13 @@ class AnimationConverter:
 
                 # Calculate the bezier interpolation nonsense
                 if fkey.interpolation == "BEZIER":
-                    keyframe.in_tans[idx] = -(value - fkey.handle_left[1])  / (frame_num - fkey.handle_left[0])  / fps / (2 * pi)
-                    keyframe.out_tans[idx] = (value - fkey.handle_right[1]) / (frame_num - fkey.handle_right[0]) / fps / (2 * pi)
+                   # beziers don't work properly so they are made to work consistently.
+                    if age.bad_beziers:
+                        keyframe.in_tans[idx] = 0.0
+                        keyframe.out_tans[idx] = 0.0
+                    else:
+                        keyframe.in_tans[idx] = -(value - fkey.handle_left[1])  / (frame_num - fkey.handle_left[0])  / fps / (2 * pi)
+                        keyframe.out_tans[idx] = (value - fkey.handle_right[1]) / (frame_num - fkey.handle_right[0]) / fps / (2 * pi)
                     bez_chans.add(idx)
                 else:
                     keyframe.in_tans[idx] = 0.0


TikiBear:
Companion pull requests: material.py, op_export.py, ui_world.py

Hoikas:
Looking back at this, I am wondering if the problem with the bezier handles is that they are using the blender frame number (arbitrary FPS) instead of the Uru frame number (fixed 30 FPS).

TikiBear:
The bezier handles can be set manually. I have tried that approach already by setting them to integers. How would they be set to match the Plasma frame numbers?

TikiBear:
One issue is that Bezier export isn't repeatable. I get different results every time I export even if I don't save the Blend file in between. It could be that every time I export the age the situation improves, but I thought I had counterexamples of that. Is there some kind of cache that Korman has that's updated even if the Blend file isn't saved?

Hoikas:
The only cache Korman uses is for textures -- animations are always reexported from scratch each run. That being said, I did some major surgery on the animation converter the other day, and results are the same each export. So, I am increasingly convinced Korman's output is correct given limitations of the Plasma engine.

Are you mixing linear and bezier keyframes in a single animation? Plasma's animation system can only support one interpolation mode per animation controller. (Danger: one animation does not necessarily result in a single controller.) While Blender lets you change from linear to bezier on a per-keyframe basis, a single bezier keyframe will trigger the entire exported animation to use bezier interpolation. The exception to this is texture offset and scale animations, which only support linear interpolation.

TikiBear:
I am not mixing linear and bezier keyframes. At this point I suspect that Plasma is the problem. The purpose of this proposed change is to aid in working around the problem, not to solve it. It isn't needed to work around the problem.
It may be that you have solved the problem if you get the same results on each export. That's not the behavior I see.

TikiBear:
In case you want to see something I put the test age Blend file I'm using (test1 ID 240) in a Box folder.
https://app.box.com/s/4woeiwyy6g8wk856djuvdulfuqs0keea
AssignedSeatingX.blend is the test age. The dynamic objects in it follow three different paths. They should be two bezier circles and a double bezier circle.

Hoikas:
I tried to export the AssignedSeatingX.blend file, but it looks like not all the textures are packed. If you can make sure they're all packed, I'll try again and see if I can find anything wrong with the exported data.

TikiBear:
I think I got all of the textures this time. Thank you!

Hoikas:
Ok, I exported the Age against what is going to be Korman 0.11, and [this is the result](https://www.youtube.com/watch?v=1RR5FmjbcYw). The only thing I notice is that the sphere speed up between the keyframes, but that matches how the animation looks in Blender's playback and is how bezier interpolation looks in general. Do you see anything in the video that I am missing?

TikiBear:
That looks correct! Maybe you've already solved the problem.

Hoikas:
Maybe so - I'm still not 100% sure what the problem was. Korman 0.11 will (hopefully) come out in a few days with a lot of bug fixes. I'm going to close this PR, but please let me know if you see anything else that doesn't look right when it comes out. The animation system still needs some significant work to support NLAs and armatures.

TikiBear:
Thank you! I'll test 0.11 and let you know what I discover.
Will there be animation event detectors at some point? The other file in that directory has an elevator based on Teledahn's elevator. I have work-arounds for the missing detectors.
Looking forward to 0.11!
 
 
-----Original Message-----

Hoikas:
I expect so. They sound useful, but I've got quite a bit of design work to do around getting multiple animations per object and material to make sense (and support armatures). No promises on when it will get done.

TikiBear:
I have loaded Korman 0.11. Yay! I have some observations already.

Animation still doesn't always work correctly. Oh, well. That can be worked-around to some extent. For example, AssignedSeating has one sphere orbiting with separate X and Y animations in quadrature. As long as all keyframes have tangents=zero the exported animation will always work correctly.

Objects with animation (whether or not it's active) and objects that are kickables don't take light at all if their material's light group contains any static lights. Not a showstopper, but light groups will have to be rearranged.
I exported the AssignedSeating file and the animated objects were entirely black. I had to remove the animation, not just deactivate it, to make them illuminated. Then, I fixed the material's light group and everybody lit up.

In Sprite Gallery, kickables were black. Objects with animation were black. Objects that have animated parents, but no animation of their own, were illuminated. I found no way to make kickables or animated items lit in Sprite Gallery. Even removing the material light maps entirely leaves these items black.

I'll let you know what else I find.
 
 
-----Original Message-----

Hoikas:
Any object that can move in game should no longer receive static (baked) lighting unless you force the issue with the lighting modifier. The movement means that the baked lighting will be wrong for every frame except frame 0. So, you have to properly set up light groups with runtime lights for these objects 😉

TikiBear:
I am unable to make lightmap texture baking work. Even in 0.11a. Vertex colors don't bake correctly either. Is there a new setting of some kind I need?
 
 
-----Original Message-----

DoobesURU:
Could you elaborate a bit more? How are they not baking correctly? Are they too bright? Too dark? Shadowing wrong? Full bright?

boblishman:
I have the same problems Doobes. I made a video to show you the difference between Korman 0.10 and O.11a.
Animated objects with vertex painting just become black in Korman 0.11a.

It has also screwed my wavset colours.

Here's a link to a video that will show you exactly what has happened. 

https://drive.google.com/file/d/1NcqN-gveL1BzTw616yZqd0_skakx3VVE/view?usp=sharing

DoobesURU:
OK, any animated objects or kickables need runtime lighting. I've noticed that even Cyan's Ages have this too.

To do this, you need to:

* Select your clickable
* Make sure any lightmap modifier is gone
* Make sure it has no Col or Alpha vertex coloring
* Add the "Lighting Info" modifier
* Check "Force RT Lighting"

This, in combo with the light group, should do the trick for any animated objects and kickables. If things show up too bright, you can adjust the diffuse color. If it's too dark, the lamps in the grouping may need adjusting or it still has a vertex color (Col) value that's overlapping.

TikiBear:
Indeed. I can certainly say that shadows aren't correct for lightmap textures. The brightness varies, but it's sort of random rather than being related to what would be a shadow.

For vertex colors I have transparent foliage that self-cast shadows that made deeper layers darker than the ones on top. Now there seems little shading effect at all.
I gave ametist the Blend file for the fanciest shadow effects. I think that she hasn't migrated so she could make correct age files for you to see as well as the Blend file itself. I'll ask her to see if she can make these available.
 
 
 
-----Original Message-----

TikiBear:
I wasn't able to get ametist's file to run on URU:CC. While she looks into that I continued to examine what I was seeing.
For the lightmap things looked really weird because the material I was looking at has multiple textures and a blend. Only the lowest texture was being lit, the other layer dark. When I reduce the material to a single texture the whole material is lit without any apparent shadows or shading (full bright?). Additional weirdness is that shading on other objects flashes on and off depending on the avatar's orientation.

I was able to get vertex lighting to work. Maybe I had an unpalatable setting, earlier, that I hadn't taken into account.

I'll let you know if I make any other discoveries.
 
-----Original Message-----

TikiBear:
Here is the 0.10-exported age package, kindly prepared by ametist. It is going to be a new age, The Fallen, so it has its own, official ID. Of interest are the foliage of the big trees, which are baked vertex light and the floor of the age, which is a baked lightmap texture.

https://www.dropbox.com/s/prpqgt89k41hn10/TheFallenExport.zip?dl=0

Here is the Blend file.
https://app.box.com/s/suo2yua394rxn3gdesgrhas66okjlisz

 FYI mostly the items are proof-of-concept. Not the way the final age would be set up.
 
 Thank you for your help!
-----Original Message-----

TikiBear:
Regarding the Korman 0.10 export I should say that the foliage in the big trees does look correct. Its light is from vertices. The ground is a light map and that is what's wacky.

Thank you!
 
 
-----Original Message-----

TikiBear:
Shorah, Adam,

Here's a question,
Node trees are sensitive to wire ordering. I've had to reorder wires to get things to work, but without any kind of guide. Is there an explanation of how to do that somewhere? Have I missed that?
TikiBear
 
 
-----Original Message-----

Hoikas:
I noticed that as well in #205, and the wire order issues should be fixed in Korman 0.11. Are you still seeing any place where the nodes are sensitive to wire ordering? To me, having to draw the wires in a specific order is a bug 😞

TikiBear:
Adam,

I still have to draw wires in a specific order with Korman 0.11a for URU:CC (where I do most of my testing). Often thingswork in URU:CC, but not DI. I have just reordered some wiring for that reason. I have yet to see if it works in DI. Diafero hasbeen patient in uploading my attempts.
The example I'm looking at now that fails in URU:CC is part of the elaborate elevator node tree. I have already uploaded that Blend file. The DI upload has already been corrected. Here is the uncorrected Blend file:

https://app.box.com/s/q9d7o1odc2vmb1vg1sxebc3eykv6yvi6
The offender is part of the Controller0 node tree. It's in the upper left had corner on the responder labelled "V1 3 to 2" ('vator 1 floor 3 to floor 2). The sound connected to thetime delay is on the bottom. I need to put it on top for the animation on the adjacent wires to work reliably in URU:CC.
 
 The change whose results I'm waiting on that will soon be in DI is for the V1Evt1 node tree. It's one of the faux animation event detectors. Its sensor node is on the top and it won't fire. The others have the sensor on the bottom. The version I have uploaded to DI has all those sensors on the bottom. We'll see if that works. If you'd like something simpler I can probably make something.
Thank you for looking into this!

Tiki

-----Original Message-----

TikiBear:
Oops. The uploaded Blend file already had the first issue corrected. I thought it had not. So, the wires on responder "V1 3 to 2" are correct in that file. Sorry for the mixup.
 
 
-----Original Message-----

Hoikas:
The Controller0 node tree looks like the bug I fixed in #205. I found this when building my sample puzzle Age... I was able to work around it by removing every wire and connecting the nodes in a specific order, but I don't remember what that order was. Then, I used that to write the fix. I haven't done any testing on Controller0 yet, so I will have to get back to you on that.

RE: V1Evt1... DI seems to have issues with regions triggering. I think ametist has specifically mentioned on the forum that some sound and footstep regions don't behave correctly in her Ages on DI but work fine offline. I think this indicates that either DI's server (Alcugs) or the PotS client have bugs with trigger regions when used online. If Korman were at fault, then I'd expect for it to be broken both offline and online. FWIW, it shouldn't matter which socket you link your region sensor to on the Python File node. Korman will export the same thing in either case -- responders are a little more complicated with their resumable nature, hence the bugs there.

TikiBear:
Adam,

I have inoperable footstep regions in DI as well and have tried a variety of shenanigans to get around that. No success, yet.

My latest update may do nothing for region triggering. I was hoping that, while DI is a fault WRT triggering, I might be able to appease it with wiring shenanigans. Neither ametist nor I have gotten the footstep regions to work, so maybe that's a lost cause at the moment. I'll let you know what happens with the wiring tweak, at least. Or, failing that, I'll let you know if I can make the sensor fire in another way.

The Controller0 tree does work in both DI and URU:CC once I rearranged the wires, if that wasn't clear. I thought the Blender file I uploaded had the "bad" version, but I was mistaken.

With lightmaps not working I've tried baking shadows to vertex colors. That works, but if I have a lamp to give the avatar an RT shadow then the baked shadows don't appear. I may never have tried that before. It's just a data point.

Thank you for your help!
TikiBear
 
 
-----Original Message-----

TikiBear:
Shorah, Adam,
This has ended up having a list of items.

Another curiosity I'm seeing is being unable to get transparent textures, that use the ADD blend, to work. I used that to reproduce CYAN's hot, crisp nimbus effects around lights. Now I can't make them visible at all, even if I actually make them opaque. MIX blends still work.
It occurs to me only now to try giving those textures alpha equal to one. Previously I had to remove the alpha channel to get the effect to work.

I address the nimbus swivels in a different pull request. The swivels added by my update are OK. Indeed all my remaining updates, and one I have yet to submit, work in Korman 0.11a.
The extra update is giving exclusion regions multiple safe targets. The logic is already present to send the avatar to the nearest safe target and I wanted to make use of that. A single region can hurl multiple avatars in all directions, as with Sprite Gallery's door force field. Maybe I'll submit it for the next Hacktoberfest. Or, maybe there's already a way to do that I don't know about.
I have been able to manually bake lights to vertex colors and give the avatar an RT shadow. Previously that could be done with the bake modifier.
Thank you for looking into Korman's mysteries!
TikiBear
 
 
-----Original Message-----

TikiBear:
Shorah, Adam,
Giving the nimbus textures unit alpha did make them visible. Yay!
Further tweaks got me where I wanted to go with vertex color light baking and RT avatar shadows.
Still no success with lightmap textures.
Thank you for your help!
TikiBear
 
 
-----Original Message-----

TikiBear:
Shorah, Adam!
Indeed, my efforts to fix the elevator in the Chamassa Restoration have failed for DI. It's fine in PoTS so Diafero has put it on UAM. DI work aroud: If you walk along the wall past the book plinth a secret sensor will reawaken an inert elevator.

I was reminded of some randomness: The Restoration animation is discontinuous. Yet, if you are using the third-person camera it discontinuously moves with the avatar and animation. This is because time step 0 is not used. I found that if I used time step 0 the third person camera had to chase the avatar. It is a curiosity.
Thank you for your help!
Your TikiBear
 
 
-----Original Message-----

TikiBear:
Adam,

Another oddity: the Auto Smooth mesh attribute doesn't affect export now. Some of my ages' geometry suddenly became child-safe (no sharp edges). The Edge Split modifier still works, though.
 
 Mark
 
-----Original Message-----

TikiBear:
Adam,
I tried to reproduce the Auto Smooth problem outside of its original context without success. Oh, well,
TikiBear
 
 
 
-----Original Message-----

TikiBear:
Shorah, Adam!
BTW, I had to modify Korman 0.11a to get shadow-only lamps. Are you addressing this?
 
 TikiBear
 
-----Original Message-----

PR #224, opened 2020-10-12
Special Export Features: Update op_export.py

TikiBear:
Add enabling check boxes for texture-map-on-export and hobble-beziers-on-export features.

### DIFF
--- a/korman/operators/op_export.py
+++ b/korman/operators/op_export.py
@@ -63,6 +63,14 @@ class PlasmaAgeExportOperator(ExportOperator, bpy.types.Operator):
                                         "description": "Forces the Blender System Console open during the export",
                                         "default": True}),
 
+        "transform_offsets": (BoolProperty, {"name": "Transform Texture Offsets",
+                                        "description": "Transforms Blender Texture Offsets to Plasma during the export",
+                                        "default": False}),
+
+        "bad_beziers": (BoolProperty, {"name": "Hobble Bezier Animation",
+                                        "description": "Forces Worst-Case Beziers during the export",
+                                        "default": False}),
+
         "texcache_path": (StringProperty, {"name": "Texture Cache Path",
                                            "description": "Texture Cache Filepath"}),
 


TikiBear:
Companion pull requests: animation.py, material.py, ui_world.py

TikiBear:
Texture offset issue: Blender's texture coordinates are different from Plasma's. The first of the two proposed features maps the coordinates on export.
Bezier issue: The tangents in bezier fcurves are randomly set to zero on export. The second proposed feature sets them all to zero so results are repeatable.
I don't expect these features to be of particular interest since everyone must already take the issues into account in their own way. But, it's hacktoberfest so here are my approaches.
If there's actually a fix for the bezier issue please let me know!

Hoikas:
I'm not intending to ignore these pull requests - I've just been occupied with lots of projects and am still trying to wrap my head around this particular issue.

Hoikas:
Regarding braziers, I've checked in with @DoobesURU to see if he's noticed anything odd. Are you exporting with the FPS set to 30? It's possible that either we have a math error in the exporter for non-30 fps values or we are running into another integer math truncation issue in the engine. Using non-30 FPS values is a really bad idea, especially for the MOUL engine, which only uses integer frame numbers for its animations.

TikiBear:
I always export with FPS=30. I'll double check, though.

TikiBear:
I checked and am exporting at 30FPS.
Integer truncation is an issue I hadn't considered. I'll look into that.

TikiBear:
I tried some tests and see an interesting result. Results haven't been repeatable, though, so I'll try some more things.

TikiBear:
I continue to get random zero tangents. I have tried in the past to figure out a pattern, but I haven't found one yet.

TikiBear:
When I say random I mean that every time the age is exported I get a different trajectory even if I don't change the animation. I wonder if Blender's dodgy memory management is the culprit.

Hoikas:
> a different trajectory

Does this mean the animation itself is different or the InTan and OutTan in PRP Shop are different?

TikiBear:
The observed animation is different. I haven't used PRP shop, but the way to duplicate the strange trajectories is to set InTan and OutTan in Blender to zero.

TikiBear:
I used PRPShop to look at the animation XML object. The InTan and OutTan values appear to be correct, counter to the observed trajectory. I don't know where to look next.

TikiBear:
Note that the bezier hobbling feature isn't required to make animation that works around the bug. It just removes the random element and gives consistent feedback.

Hoikas:
I wonder if this is related to the issue with incorrect coordinates due to object parenting that we see in subworlds. :/

TikiBear:
The trajectories match at the keyframes, so I don't think it's related to subworld parenting. Only the movement between keyframes is incorrect.

TikiBear:
I propose this be closed. I think it's really overkill at this point. Note that it contains something for #221, which is closed IFO #243.

PR #226, opened 2020-10-20
Fix Exponential Fog

DoobesURU:
This was something minor that came up awhile back. For some reason, SetDefExp2 type fog doesn't seem to work upon export in either MOUL or PotS/CC, but regular SetDefExp does. This is just a proposed minor change to use the regular SetDefExp option if fixing the other is too big a task.

Yeah, could probably have left everything except `Graphics.Renderer.Fog.SetDefExp` alone, but I figured we'd want things uniform across the board. 😉

### DIFF
--- a/korman/properties/prop_world.py
+++ b/korman/properties/prop_world.py
@@ -31,7 +31,7 @@ class PlasmaFni(bpy.types.PropertyGroup):
     fog_method = EnumProperty(name="Fog Type",
                               items=[
                                      ("linear", "Linear", "Linear Fog"),
-                                     ("exp2", "Exponential", "Exponential Fog"),
+                                     ("exp", "Exponential", "Exponential Fog"),
                                      ("none", "None", "Use fog from the previous age")
                                     ])
     fog_start = FloatProperty(name="Start",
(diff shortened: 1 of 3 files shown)


korman/properties/prop_world.py
```diff
@@ -31,7 +31,7 @@ class PlasmaFni(bpy.types.PropertyGroup):
     fog_method = EnumProperty(name="Fog Type",
                               items=[
                                      ("linear", "Linear", "Linear Fog"),
-                                     ("exp2", "Exponential", "Exponential Fog"),
+                                     ("exp", "Exponential", "Exponential Fog"),
```
Hoikas, line 34: This change *may* impact existing blend files.

DoobesURU: This was actually tested with an existing blend...in this case, Kotsahvosahn, the waystation. Seemed to export fine and didn't seem to affect it negatively AFAICT.

Hoikas: The specific case I'm worried about is if blends that are already set to "exp2" suddenly change to linear or none when opened with this change. Was the blend you tested already set to "exp2" and correctly converted to "exp"?

DoobesURU: Er, yes...it randomly fogs certain items in linear and exponential stays stuck at its FNI value even with console modification. "None" still seems to do what it's supposed to.

And here I thought this'd be a small change. Ha!

DoobesURU: So an update...I see why the fog was off...and it wasn't due to the change, but it may be something to address.

The start fog value was larger than the end fog value, which is what was causing the issues above. I switched the values and it works as it should now...although not related, linear fog looks weird for the waystation.

I tried adding a few if/else lines to find this exception and raise an error, but...it crashed Korman and Blender completely. 😮 

Would it be worth raising an error if this happens, and if so, what the best, non-crash way to do so?

Issue #228, opened 2020-10-29
Lighting Problems Since PR #210


DoobesURU:
Working on the Pub's holiday additions, I have noticed a few things since #210:

- Having Shadeless checked on a mesh set with a modifier for a lightmap...and occasionally on a vertex colored mod...doesn't produce the full bright effect and, in fact, shadows the area anyway. Also, it appears, at least when a mesh has its scale animated (in this case, the candle flames on the menorah imported from the city), that shadeless also doesn't work.

- New materials, or perhaps meshes with newly added Bake Lighting modifiers, do not light properly despite having the same settings as existing (and properly lit) meshes and materials.

- While it doesn't show in the final export, meshes with an existing Col vertex paint are once again producing an unnecessary "autocolor" Vcol layer.

I'm still checking to see if there's anything else, but this is what's stuck out so far.

DoobesURU:
Another observation: animated objects are also not lighting properly unless parented to an empty/dummy object.

DoobesURU:
Update #2: So I tried a few things:

1. It looks like there were a few instances of `plasma_modifiers.lighting.rt_lights` in the materials and mesh export scripts. Was this meant to be `plasma_modifiers.lighting.force_rt_lights` as it says in the modifier?

2. While on the modifier, it seems to also be using the force_rt_lights bool value even if the modifier itself isn't active on a particular object. Adding the mod back and unchecking Force RT Lights seems to fix this for some objects. Still testing this on others.

Hoikas:
Closing after #230

PR #229, opened 2020-11-06
More DSpan material fixes

DoobesURU:
Fixes a few material errors that cropped up after PR #210  and detailed in issue #228 , but could do with some more checking.

Changes include:
- sanity checks to make sure the Lighting Info modifier is actually enabled. This is because if the mod had been used, then deleted, Korman still seems to read the bool values anyway.
- changed instances of `lighting.rt_lights` to `lighting.force_rt_lights`.
- Exception for runtime color and wavesets. The runtime must be black in that case or the waveset is too bright.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -188,7 +188,7 @@ class MaterialConverter:
             hgmat = None
         else:
             # Ensure that RT-lit objects don't infect the static-lit objects.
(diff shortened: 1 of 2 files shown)


DoobesURU:
Closing since Hoikas' #230 did a more sane job of fixing this.

PR #230, opened 2021-01-06
More DSpan Lighting Fixes

Hoikas:
This is an attempt to address some more issues with runtime lighting, Particularly, we now force any light attached to a PFM to be `kLPMovable` - because the only reason you would attach a light to a PFM is to either warp it or turn it on/off. Further, we now export the layer runtime color as the material diffuse unless preshading is explicitly off. The behavior change is subtle, but we only want to set runtime to black in the case of purely dynamic lighting. Otherwise, we end up restricting receiving runtime lighting to only objects that are `kLiteVtxNonPreshaded` because `kLiteMaterial` multiples the runtime color by the lamp color.

### DIFF
--- a/korman/exporter/manager.py
+++ b/korman/exporter/manager.py
@@ -16,6 +16,7 @@
 import bpy
 from pathlib import Path
 from PyHSPlasma import *
+from typing import Iterable
 import weakref
 
 from . import explosions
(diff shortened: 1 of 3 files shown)


dpogue (approved):
Looks okay to me

PR #231, opened 2021-02-05
Wind Object fix

DoobesURU:
A long time coming but only now just fixed, this makes a small correction to the wind object part of the waveset modifier to get it to work.


Hoikas (approved):
Whoops

PR #236, opened 2021-02-19
Implement Material Diffuse Animations

Hoikas:


DoobesURU:
Several Age exports and animations seem to still be working. Animated the diffuse color of one of the calendar sparks and that's working as well.

PR #242, opened 2021-04-15
Region sensor: assume receiver on parent_so instead of own sceneobject

Jrius:
Just like clickable nodes, region sensor nodes now assume their receivers to be attached to the parent_so (the object with the Advanced Logic modifier) instead of attached to the region itself.

Previously, a region sensor would instead create its own new empty receiver (say, a Responder with zero states) and notify it instead of the real responder, rendering it useless.

### DIFF
--- a/korman/nodes/node_conditions.py
+++ b/korman/nodes/node_conditions.py
@@ -432,21 +432,21 @@ class PlasmaVolumeSensorNode(idprops.IDPropObjectMixin, PlasmaNodeBase, bpy.type
                                            member_group="kGroupDetector",
                                            report_groups=self.report_on)
 
-    def _export_volume_event(self, exporter, bo, so, parent_so, event, settings):
+    def _export_volume_event(self, exporter, region_bo, region_so, parent_so, event, settings):
         if event == plVolumeSensorConditionalObject.kTypeEnter:
             suffix = "Enter"
         else:
             suffix = "Exit"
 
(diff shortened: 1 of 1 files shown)


korman/nodes/node_conditions.py
```diff
@@ -417,22 +417,22 @@ def export(self, exporter, bo, parent_so):
...
+            key = self._export_volume_event(exporter, region_bo, region_so, parent_so, plVolumeSensorConditionalObject.kTypeExit, exit_settings)
             interface.addIntfKey(key)
 
         # Don't forget to export the physical object itself!
         exporter.physics.generate_physical(region_bo, region_so, bounds=self.bounds,
                                            member_group="kGroupDetector",
                                            report_groups=self.report_on)
 
-    def _export_volume_event(self, exporter, bo, so, event, settings):
+    def _export_volume_event(self, exporter, bo, so, parent_so, event, settings):
```
Hoikas, line 435: Now that we are juggling two different scene objects, I think we should rename `bo` and `so` to be more descriptive.

Hoikas (commented):
This looks right, but I want to come back and read it a little bit later to make sure. I'm sure you understand that getting this kind of thing right is hard :smile:

Jrius:
Haha ! Sure, no problem, that bug was a bit hard to explain. (Better get ready, I'm working on some "fun" lightmapper stuff for another PR :P )
Variable names, should have thought about it earlier... Done !

Hoikas (approved):
Thanks for the fix!

PR #243, opened 2021-04-16
Convert layer offsets to Plasma's space

Jrius:
Blender's texture slot scaling is done from the center (0.5, 0.5) of the UV mapping viewport, with Y axis up. Plasma's scaling is done from origin (0, 0), with Y axis down. Ah, D3D and OpenGL...

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -420,7 +420,7 @@ class AnimationConverter:
             scale = kwargs[scale_path]
 
             translation = hsVector3(pos[0] - (scale[0] - 1.0) / 2.0,
-                                    1.0 - pos[1] - (scale[1] - 1.0) / 2.0,
+                                    -pos[1] - (scale[1] - 1.0) / 2.0,
                                     pos[2] - (scale[2] - 1.0) / 2.0)
             matrix = hsMatrix44()
             matrix.setTranslate(translation)
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -419,8 +419,11 @@ class AnimationConverter:
             pos = kwargs[pos_path]
             scale = kwargs[scale_path]
 
+            translation = hsVector3(pos[0] - (scale[0] - 1.0) / 2.0,
+                                    1.0 - pos[1] - (scale[1] - 1.0) / 2.0,
+                                    pos[2] - (scale[2] - 1.0) / 2.0)
             matrix = hsMatrix44()
-            matrix.setTranslate(hsVector3(*pos))
+            matrix.setTranslate(translation)
             matrix.setScale(hsVector3(*scale))
             return matrix
 
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -425,7 +425,10 @@ class MaterialConverter:
 
         # Transform
         xform = hsMatrix44()
-        xform.setTranslate(hsVector3(*slot.offset))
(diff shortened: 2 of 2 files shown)


Jrius:
Wait, this is similar to #221 by TikiBear. I should have checked, sorry :/

Jrius:
Hmm, TikiBear's scope and solution seem different from mine, AFAICT. Not sure about the specifics. I'm leaving this open, feel free to close.

Sorry again, should have checked before the PR.

Hoikas:
> Wait, this is similar to #221 by TikiBear. I should have checked, sorry :/

No worries. I was really struggling with understanding what the goal of #221 was. Now I know that this is a difference between DirectX and OpenGL. I am, however, a little concerned about what this means for already created Ages. How will this affect them and how significant is this fix for actually working in Blender?

Jrius:
Importance: quite important IMHO since currently Korman exports texture offsets that don't match what one can see in Blender's textured view. We don't want that.

Repercussions on existing Ages: if texture offset/scale was untouched, none at all. Otherwise, yeah, some textures will have different offsets (which includes texture offset animation on the Y axis being reversed to match Blender's behavior). In most cases, the user will not notice unless they really rely on precise texture offsetting - which, in that (uncommon) case, they would have noticed the issue and should have reported it already.

The only case in which someone might not know about the issue and be negatively affected by this fix is if they animated some texture offset and exported without first previewing it in Blender. For instance, if they made an animated waterfall with a texture sliding on the Y axis, but entered a random offset value and didn't bother to check the direction or speed of the texture until they exported in Plasma. As unlikely as it sounds, it _could_ have happened due to Blender 2.79's absolutely horrendous playback of texture mapping animations, which some people might have thought broken. (Previewing texture animation requires either a full render with Ctrl-F12, or to pause the anim each frame and force Blender to update the viewport version of the material - usually by reentering one of the material's mapping value. Yeah, it's stupid.)
Still, in that case, fixing the animation takes just a few seconds.

Hoikas:
Ok. This sounds good, and I like that this changeset uses the built-in keyframe converter where #221 opts for some extra manual processing. I'd like for @DoobesURU to chime in on the impact to preexisting Ages. Otherwise, I'm 👍 on this.

DoobesURU:
Just letting you know I saw this notification. I'll put in my two cents as soon as I'm back in front of my computer.

Jrius:
> So I think it will only affect texture animations that aren't continuous?

Animation continuity won't make a difference, but accurate texture offset/positioning will. In your laptop case, it's hard to say what might be wrong without looking at your animation (AFAIK the laptop animation is not on either releases of the Pub that I have access to). I guess, if your texture is set to not repeat ("clip" or "extend" as mapping extension in Blender), it might cause the texture to disappear since it might now be out of UV bounds (which would NOT be visible in the Blender viewport since 2.7x cannot clamp textures correctly :angry:  ).

I'll let you run your tests but if you happen across other suspicious behavior don't hesitate to let me know and I'll investigate ;)

(Worth mentioning I'm testing in Tsoidahl Prad, which was setup for PyPRP, which [did the same as this fix](https://github.com/H-uru/PyPRP/blob/master/PyPRP/prp_MatClasses.py#L1062))

Hoikas:
@DoobesURU Just to make sure, we're good to merge this one?

DoobesURU:
> 
> 
> @DoobesURU Just to make sure, we're good to merge this one?

I believe so. Fire away, boss! 😄

PR #245, opened 2021-04-23
Allow GroupMaster (MsgForwarder) to animate object data (such as lamps)

Jrius:
Pretty much what it says. The GroupMaster modifier expects to find animation data on the `object` itself and doesn't check the `object.data`, thus preventing you from adding an animated lamp to its list. This fixes that.

### DIFF
--- a/korman/idprops.py
+++ b/korman/idprops.py
@@ -119,13 +119,7 @@ class IDPropObjectMixin(IDPropMixin):
 
 
 def poll_animated_objects(self, value):
-    if value.animation_data is not None:
-        if value.animation_data.action is not None:
-            return True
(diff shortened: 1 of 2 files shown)


Hoikas (commented):
It might be better to use the (newer) convenience property [`plasma_object.has_animation_data`](https://github.com/H-uru/korman/blob/79f81fc63aad1b4744b44766a6b1a3fd012901cb/korman/properties/prop_object.py#L60:L71) instead of reimplementing it everywhere.

Jrius:
Ah, looks much better indeed.

Hoikas (approved):

Hoikas:
TY for the fix!

PR #246, opened 2021-04-24
Fix emissive lightmaps broken by a203e10.

Hoikas:
I'm not really sure what value an emissive lightmap has, but apparently people use them. So, it's better to give not-quite-so-emissive results instead of breaking the entire Age's rendering.


Hoikas:
@DoobesURU when you get a chance, could you make sure this breaks nothing?

DoobesURU:
I've exported three Ages using this fix: Tiam, Great Tree Pub and my Pub. All seem to be OK AFAICT.

PR #247, opened 2021-04-26
Enable Shadow-only Lamps

TikiBear:
Enable shadow-only lamps.

This is what I did to enable shadow-only lamps. It might seem counter-intuitive. I know that other lighting work is underway so this could already be moot.

### DIFF
--- a/korman/exporter/rtlight.py
+++ b/korman/exporter/rtlight.py
@@ -138,14 +138,14 @@ class LightConverter:
         shadow_only = bl_light.shadow_method != "NOSHADOW" and bl_light.use_only_shadow
 
         # Apply the colors
-        if bl_light.use_diffuse and not shadow_only:
+        if bl_light.use_diffuse: # TikiBear
             self._report.msg("Diffuse: {}", diff_str, indent=2)
             pl_light.diffuse = hsColorRGBA(*diff_color)
         else:
             self._report.msg("Diffuse: OFF", indent=2)
             pl_light.diffuse = hsColorRGBA(0.0, 0.0, 0.0, energy)
 
-        if bl_light.use_specular and not shadow_only:
+        if bl_light.use_specular: # TikiBear
             self._report.msg("Specular: {}", spec_str, indent=2)
             pl_light.setProperty(plLightInfo.kLPHasSpecular, True)
             pl_light.specular = hsColorRGBA(*spec_color)
--- a/korman/exporter/rtlight.py
+++ b/korman/exporter/rtlight.py
@@ -138,14 +138,14 @@ class LightConverter:
         shadow_only = bl_light.shadow_method != "NOSHADOW" and bl_light.use_only_shadow
 
         # Apply the colors
-        if bl_light.use_diffuse: # TikiBear
+        if bl_light.use_diffuse:
             self._report.msg("Diffuse: {}", diff_str, indent=2)
             pl_light.diffuse = hsColorRGBA(*diff_color)
         else:
             self._report.msg("Diffuse: OFF", indent=2)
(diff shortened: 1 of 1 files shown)


DoobesURU:
I've been meaning to ask about shadow only lights since they seem to be a big staple of Cyan's Ages. I'm guessing these are, as the name suggests, lights that only cast shadows real-time and not any actual lighting?

TikiBear:
Yes. This is for RT lamps that provide no diffuse or specular light to the scene, just a shadow. The issue is the need to bake static lights and shadows. If the RT lamp that makes the avatar shadow also provides diffuse light it can wipe out the effects of the baked lighting. This is what I see without the fix.
 
 TikiBear
 
-----Original Message-----

Hoikas (commented):
I feel like this should be the other way around, e.g. by testing `if not shadow_only:` instead of checking `use_diffuse` and `use_specular`. Also, we prefer to avoid tagging the bounds of individual patches in favor of using git's built-in history commands (eg `git blame` and `git log`).

TikiBear:
I did have a Senior Moment (tm) with the blame command! I can redo the request, but it sounds as though you all have a different approach anyway. I'm fine just closing this.

Deledrius:
Thanks for this submission!  It seems like a useful option to have available.

By the way, there's no need to close the PR, you can (or we can) update the commits in a PR to adjust it for acceptance. :)  I was going to point out but Hoikas beat me to it: no need to comment authorship as git tracks that through each commit itself.  It's better to comment what a segment of code is doing for future reference.  If you're able to make the changes requested, let us know, otherwise we can fix the patch as necessary before accepting it.

It would be helpful though in the future if you gave the PR a more descriptive name.  It seems the default option Github does is just name the file(s) being modified, but if you can edit that it would be a lot easier to know what the intent is for your submissions.

All that said, thanks again for contributing!

TikiBear:
I thought that I couldn't change the PR title. I wonder what I'm doing wrong.

Removing the shadow_only test is, in itself, the fix. If that isn't acceptable then there's nothing to turn over.

Deledrius:
I believe @Hoikas is referring to swapping the conditional test (and subsequent code), which would still have the desired effect but be more explicit in the intent of the test.

TikiBear:
I don't understand. The shadow_only flag should not be tested at all.

TikiBear:
I have tried to tidy up my proposed changes. Let me try to clarify that the fix is the removal of the shadow_only tests.

Hoikas:
> I feel like this should be the other way around, e.g. by testing if not shadow_only: instead of checking use_diffuse and use_specular

Pinging @TikiBear for clarification about why `shadow_only` should not be tested. Per at @Deledrius, testing shadow_only would be more explicit.

TikiBear:
To make the shadow-only lamps work the light parameters (including use_diffuse) must NOT be set to zero. For example, for diffuse light, performing "pl_light.diffuse = hsColorRGBA(0.0, 0.0, 0.0, energy)" is incorrect. You still need "pl_light.diffuse = hsColorRGBA(*diff_color)". Therefore there is no need to test for shadow_only in those clauses. Apparently, Plasma casts only the shadow because of the plLightInfo.kLPShadowOnly property. The use_diffuse and use_specular flags in those clauses should allow you to create diffuse-only and specular-only lamps, though I haven't been able to get specular-only lamps to work (not that there's a big need). This is what I had to do to make shadow-only lamps work.

Hoikas:
Ah, I understand now. Thanks for the clarification. For posterity, the relevant bit is:
> To make the shadow-only lamps work the light parameters (including [the diffuse color]) must NOT be set to zero.

PR #250, opened 2021-04-29
Add Funky Blending layer support

dpogue:
### To Do
- [x] Fix the Python korlib texture generation

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -270,6 +270,11 @@ class MaterialConverter:
                             if i+1 < curr_stencils:
                                 stencil_layer.state.miscFlags |= hsGMatState.kMiscBindNext
                             hsgmat.addLayer(stencil_layer.key)
+                    if slot.texture.plasma_layer.funky_type != "FunkyNone":
+                        funky_ramp = self._export_funky_slot(bo, bm, hsgmat, slot, idx)
+                        if funky_ramp:
+                            tex_layer.state.miscFlags |= hsGMatState.kMiscBindNext
+                            hsgmat.addLayer(funky_ramp.key)
 
         # Plasma makes several assumptions that every hsGMaterial has at least one layer. If this
         # material had no Textures, we will need to initialize a default layer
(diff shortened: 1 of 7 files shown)


korman/korlib/__init__.py
```diff
@@ -29,6 +29,30 @@
...
+                start = i*j*4
+                end = start + 4
+
+                if additive:
+                    if x < y:
+                        x = y
+
+                    buf[start:end] = [b for b in (255, 255, 255, int(x * 255.9))]
+                else:
+                    buf[start:end] = [b for b in (255, 255, 255, int((x * y) * 255.9))]
```
dpogue, line 52: @Hoikas This buffer stuff is wrong, but my Python isn't great and I'm not seeing anything obvious. The C++ implementation works.

Hoikas (commented):
First pass through

korman/korlib/__init__.py
```diff
@@ -29,6 +29,30 @@
...
+
+        mipmap.setRawImage(bytes(buf))
```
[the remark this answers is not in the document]
dpogue: `setRawImage` is what's available in our PyHSPlasma Python API 🤷🏼‍♂️ 
It calls the C++ `setImageData` internally. This is consistent with how the bumpmap LUT texture is handled.

dpogue (commented):
I've rebased this, and I think I've addressed the remaining feedback

korman/exporter/material.py
```diff
@@ -377,6 +382,87 @@ def export_waveset_material(self, bo, bm):
...
+            if funky_type == "FunkyDist":
+                ramp_layer.UVWSrc = plLayerInterface.kUVWPosition
+                ramp_layer.state.miscFlags  |= hsGMatState.kMiscNoShadowAlpha
+            elif funky_type == "FunkyNormal":
+                ramp_layer.UVWSrc = plLayerInterface.kUVWNormal
+            elif funky_type == "FunkyUp":
+                ramp_layer.UVWSrc = plLayerInterface.kUVWNormal
+                ramp_layer.state.miscFlags |= hsGMatState.kMiscOrthoProjection
+            elif funky_type == "FunkyReflect":
+                ramp_layer.UVWSrc = plLayerInterface.kUVWReflect
```
Hoikas, line 463:
```suggestion
            elif funky_type == "FunkyReflect":
                ramp_layer.UVWSrc = plLayerInterface.kUVWReflect
            else:
                raise ValueError(funky_type)
```

korman/exporter/material.py
```diff
@@ -377,6 +382,87 @@ def export_waveset_material(self, bo, bm):
...
+            funky_type = layer_props.funky_type
+
+            near_trans = layer_props.funky_near_trans
+            near_opaq = layer_props.funky_near_opaq
+            far_trans = layer_props.funky_far_trans
+            far_opaq = layer_props.funky_far_opaq
+
+            if near_trans > far_trans:
+                near_trans, far_trans = far_trans, near_trans
+                near_opaq, far_opaq = far_opaq, near_opaq
```
Hoikas, line 401: It looks like it's excepted that `far_opaq` could be less than `near_opaq` and not trigger this swap based on the `additive` calculation?

korman/korlib/__init__.py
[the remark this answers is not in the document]
Hoikas: Sounds good. I probably wasn't looking at the API closely when I asked.

korman/ui/ui_texture.py
```diff
@@ -130,3 +130,28 @@ def _has_animation_data(cls, context):
...
         ui_anim.draw_multi_animation(self.layout, "texture", context.texture.plasma_layer,
                                      "subanimations", use_box=True)
+
+
+class PlasmaFunkyLayerPanel(TextureButtonsPanel, bpy.types.Panel):
+    bl_label = "Plasma Funky Layer"
+
+    def draw(self, context):
+        texture, slot = context.texture, getattr(context, "texture_slot", None)
+        use_stencil = slot.use_stencil if slot is not None else False
```
Hoikas:
```suggestion
        texture = context.texture
```

You don't seem to be using the texture slot or the stencilling info?

Hoikas (commented):
I haven't actually written this down anywhere, but I'm starting to adopt a pattern with if statements in Korman where if there is ever an `elif`, then there must be an `else:`. Even if it winds up being:
```python
else:
    pass
```
just so it's clear what the logic is. Could you go through your `@property`s and make sure there is always an `else:` if there is an `elif`?

korman/exporter/material.py
dpogue: This matches the implementation in PlasmaMax, but I copied over the comment: https://github.com/H-uru/Plasma/blob/58c833de3987fada154d914db769838ac92a6478/Sources/MaxPlugin/MaxConvert/hsMaterialConverter.cpp#L2962-L2967

Hoikas:
This is starting to look good. Before I merge, a couple of last questions.
- The exact appearance of these effects isn't super clear to me or what all the knobs do. Do you have a test blend that shows the feature in action? This will help when I go to write the release notes.
- It looks like in PlasmaMAX, the funky distances can feed into the span's min and max distances. Any thoughts about whether or not that should be supported immediately, or it we're fine to defer on that?

PR #251, opened 2021-05-03
Implement the Random Sound modifier.

### LINKED ISSUE #165: Random Sound Modifier
Need to expose `plRandomSoundMod` for random background music stingers and kickable collision sound effects.

Hoikas:
Fixes #165
Closes #212
Depends on H-uru/libhsplasma#206

[Sample Age](https://github.com/H-uru/korman/files/6412755/KormanRndSnd.zip)

TODO
- [x] Add support for enable/disable random sounds to logic nodes (RSMs expect `plAnimCmdMsg`, not `plEnableMsg`)

### DIFF
--- a/korman/properties/modifiers/sound.py
+++ b/korman/properties/modifiers/sound.py
@@ -22,8 +22,132 @@ from PyHSPlasma import *
 
 from ... import korlib
 from .base import PlasmaModifierProperties
+from .physics import surface_types
 from ...exporter import ExportError
 from ... import idprops
+
+_randomsound_modes = {
+    "normal": plRandomSoundMod.kNormal,
+    "norepeat": plRandomSoundMod.kNoRepeats,
+    "coverall": plRandomSoundMod.kCoverall | plRandomSoundMod.kNoRepeats,
+    "sequential": plRandomSoundMod.kSequential
+}
+
+class PlasmaRandomSound(PlasmaModifierProperties):
+    pl_id = "random_sound"
+    pl_depends = {"soundemit"}
+
+    bl_category = "Logic"
+    bl_label = "Random Sound"
+    bl_description = ""
+
+    mode = EnumProperty(name="Mode",
+                        description="Playback Type",
+                        items=[("random", "Random Time", "Plays a random sound from the emitter at a random time"),
+                               ("collision", "Collision Surface", "Plays a random sound when the object's parent collides")],
+                        default="random",
+                        options=set())
+
+    # Physical (read: collision) sounds
+    play_on = EnumProperty(name="Play On",
+                           description="Play sounds on this collision event",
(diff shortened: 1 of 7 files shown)


DoobesURU:
Installing and giving it a try now. I'll report back after a bit of testing.

korman/properties/modifiers/sound.py
```diff
@@ -22,9 +22,133 @@
...
+                            options={"ENUM_FLAG"})
+
+    # Timed random sounds
+    auto_start = BoolProperty(name="Auto Start",
+                              description="Start playing when the Age loads",
+                              default=True,
+                              options=set())
+    play_mode = EnumProperty(name="Play Mode",
+                             description="",
+                             items=[("normal", "Any", "Plays any attached sound"),
```
DoobesURU, line 69: Would it be too much/too confusing to add the mode numbers after the Mode names (e.g. "Any (16)", "No Repeat (17)", "Full Set (19)", "Sequential (48)"? It could be helpful for anyone converting Cyan Ages and looking at the PRPs.

If it's too much, we could just leave it to future wiki documentation.

Hoikas: I think it might be better to add support for RSMs to ZLZ

DoobesURU: That'd work too.

Jrius: > add support for RSMs to ZLZ

Sure, this (and a bunch of other things) are planned for when I jump back to ZLZ. I'm always interested in supporting more Korman features. Having the user deal with the underlying numbers 'n stuff is always undesirable.

I was planning on doing some other stuff before that, but I can prioritize this if you need it Doobes.

DoobesURU: @Jrius If it's not too difficult, of course. It'd be one less thing to have to wire when converting.

DoobesURU:
I can confirm (before my power blinked) that the sound message node now works for random sound mods.

korman/properties/modifiers/sound.py
Jrius: 'K, support added to ZLZ. This also required the object's collider to parent all other objects (drawable, sounds etc), which is better behavior anyway.

Observation: Relto firemarbles use collision sound group kUser1 (great naming...), which is not available to Korman and therefore ignored. Dunno if you want to support it.

Hoikas: Oh... of course Cyan used those groups. 🙄

Deledrius:
Does this require further testing?

Hoikas:
I don't think so, but I do need to add the silly user kickable types into the enum [because Cyan](https://github.com/H-uru/korman/pull/251#discussion_r629494246).

PR #253, opened 2021-05-30
Eax reverb

Jrius:
Found a way to reenable EAX on CC the other day, I thought why not add it to Korman too.

Adds the Reverb Region Modifier (aka plEAXListenerMod), and allows enabling reverberation for some sounds. Hear how it sounds in [this video](https://youtu.be/Nxme32iVZY4).

Relies on [this HSPlasma pull request](https://github.com/H-uru/libhsplasma/pull/209).

AFAIK EAX is buggy in MOULa currently ? so requires a modded CC install to test. More info [here](https://community.pcgamingwiki.com/files/file/2149-eax-fix-for-uru-complete-chroniclesages-beyond-myst/), and you'll also need to install [DSOAL](https://community.pcgamingwiki.com/files/file/1363-dsoal/). If Windows prevents `dsound.dll` from overriding the base one, follow [this link](https://www.indirectsound.com/registryIssues.html). I'll probably write a better tutorial at the GoW at some point.

The Reverb Region Modifier is attached to a softvolume. It provides reverberation for sound sources with EAX enabled - such as footstep sounds. For ease-of-use, the user simply has to pick a reverb profile within a list of existing presets. There are several dozens of those, which are split into two lists(*). The user can also setup the reverb profile himself, if he's stupid enough to try his hand with undocumented stuff.

(*) I kept both separate since historically the first (shorter) list of presets comes from the original EAX library. The second (longer) list comes from OpenAL, if I'm not mistaken. The first list is commonly found in other game engines and softwares, the second one less so.

Sound emitters themselves now have a toggle to enable reverberation. This only enables the `plSound::plEAXSourceSettings`. (There are more parameters available in the plEAXSourceSettings, but I couldn't figure out how they worked.)

Hope you like the modifier's icon, otherwise we can change it :P


korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,237 @@ def requires_actor(self):
...
+                                 ("UNDERWATER", "Underwater", ""),
+                                 ("DRUGGED", "Drugged", ""),
+                                 ("DIZZY", "Drizzy", ""),
+                                 ("PSYCHOTIC", "Psychotic", ""),
+                                 ("MORE", "More choices...", ""),
+                                 ("CUSTOM", "Custom", "Setup your own environment")],
+                          default="GENERIC",
+                          options=set())
+
+    # Thikk list for annoying users.
```
Hoikas, line 257: For correct spelling:
```suggestion
    # Thick list for annoying users.
```

For innuendo:
```suggestion
    # Thicc list for annoying users.
```

😛

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,237 @@ def requires_actor(self):
...
+    modulation_time = FloatProperty(name="Modulation Time", description="Modulation Time",
+                                    default=0.25, min=0.1, max=5.0)
+    modulation_depth = FloatProperty(name="Modulation Depth", description="Modulation Depth",
+                                     default=0.0, min=0.0, max=1.0)
+    air_absorption_hf = FloatProperty(name="Air Absorption HF", description="Air Absorption High Frequency",
+                                      default=-5.0, min=-10.0, max=0.0)
+    hf_reference = FloatProperty(name="HF reference", description="High Frequency Reference",
+                                 default=5000.0, min=1000.0, max=20000.0)
+    lf_reference = FloatProperty(name="LF reference", description="Low Frequency Reference",
+                                 default=250.0, min=20.0, max=1000.0)
```
Hoikas, line 387: These all need an `options=set()` to prevent folks from adding spurious keyframes.

korman/properties/modifiers/sound.py
```diff
@@ -152,6 +152,16 @@ def _update_name(self, context=None):
                          options={"ANIMATABLE"},
                          subtype="PERCENTAGE")
 
+    use_reverb = BoolProperty(name="Reverberation",
+                              description="Allow the sound to be affected by reverberation regions",
+                              default=False,
+                              options=set())
```
Hoikas, line 158: I feel like this should probably be implicit. We almost definitely want all 3D sounds (read: sound effects) to be affected by environmental audio, so that's an extra button press required for them. Whereas background music, GUI sounds, ambiance, etc. probably don't want reverb. If someone wants to explicitly disable reverb on a sound effect, we could test for the `reverb_amount == 0` and just toggle it off.

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,237 @@ def requires_actor(self):
...
+                                 ("MOUNTAINS", "Mountains", ""),
+                                 ("QUARRY", "Quarry", ""),
+                                 ("PLAIN", "Plain", ""),
+                                 ("PARKINGLOT", "Parking lot", ""),
+                                 ("SEWERPIPE", "Sewer pipe", ""),
+                                 ("UNDERWATER", "Underwater", ""),
+                                 ("DRUGGED", "Drugged", ""),
+                                 ("DIZZY", "Drizzy", ""),
+                                 ("PSYCHOTIC", "Psychotic", ""),
+                                 ("MORE", "More choices...", ""),
```
Hoikas, line 252: I think it might be better to use a checkbox (read: BoolProperty) somewhere to control displaying more options in the same drop-down.

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,237 @@ def requires_actor(self):
...
+            props.echoDepth = self.echo_depth
+            props.modulationTime = self.modulation_time
+            props.modulationDepth = self.modulation_depth
+            props.airAbsorptionHF = self.air_absorption_hf
+            props.hfReference = self.hf_reference
+            props.lfReference = self.lf_reference
+            flags = 0
+            for flag in self.flags:
+                flags |= getattr(EAXReverbProperties, flag)
+            props.flags = flags
```
Hoikas, line 429:
```suggestion
            for flag in self.flags:
                props.flags |= getattr(EAXReverbProperties, flag)
```

Hoikas (commented):
Great demo! I've missed having reverb effects in Uru, and this will definitely be great for fan Ages. We *may* defer merging this until proper support is available in H-uru/Plasma - just so we don't have to point to any DLL hacking when the feature is released :wink:

Deledrius:
> 
> 
> Great demo! I've missed having reverb effects in Uru, and this will definitely be great for fan Ages. We _may_ defer merging this until proper support is available in H-uru/Plasma - just so we don't have to point to any DLL hacking when the feature is released 😉

I agree; this looks good, but I'd like to be able to test it against H'uru Plasma as well as Uru:CC if possible.

Jrius:
> 
> 
> Great demo! I've missed having reverb effects in Uru, and this will definitely be great for fan Ages. We _may_ defer merging this until proper support is available in H-uru/Plasma - just so we don't have to point to any DLL hacking when the feature is released wink

Thanks ! Sure, no problem, EAX is a nice feature but it's not a priority.
The DLL wrapper will still be required on CC, but someday I'll see if Drizzle/UAM can install it automatically.

> 
> 
> I agree; this looks good, but I'd like to be able to test it against H'uru Plasma as well as Uru:CC if possible.

Sure. Let me know if you want the Tsoidahl Prad source to test, I can probably upload it somewhere.
(I'm a bit curious to know if there are any difference between the original EAX library, DSOAL and the upcoming OAL implementation.)

korman/properties/modifiers/region.py
Jrius: Ah, the silly version is the one I had in mind. Assuming you don't mind :P

korman/properties/modifiers/sound.py
Jrius: Makes sense.

korman/properties/modifiers/region.py
Jrius: As in, a preference in the Blender user settings (next to the game/python configuration) ? I fear most users will never find it and assume only the first part of the list is available, and it might affect sharing Blender files across installs. Maybe having both lists merged into one by default would be better ? The way Blender displays lists should make it readable enough.

I'm fine with either solution, though. Let me know what you prefer.

Hoikas: Sorry about neglecting to respond. I think merging them all into one list would be fine.

korman/ui/modifiers/sound.py
```diff
@@ -83,6 +83,10 @@ def soundemit(modifier, layout, context):
         col.separator()
         _draw_fade_ui(sound.fade_out, col, "Fade Out:")
 
+        if sound.sfx_type in ("kSoundFX", "kNPCVoices"):
```
Hoikas, line 86:
```suggestion
        if sound.sfx_type in {"kSoundFX", "kNPCVoices"}:
```

korman/properties/modifiers/sound.py
```diff
@@ -290,6 +296,21 @@ def _convert_sound(self, exporter, so, pClass, wavHeader, dataSize, channel=None
         else:
             sound.channel = plWin32Sound.kRightChannel
 
+        # Reverb/EAX
+        if self.sfx_type in ("kSoundFX", "kNPCVoices") and self.reverb_amount > 0:
```
Hoikas, line 300:
```suggestion
        if self.sfx_type in {"kSoundFX", "kNPCVoices"} and self.reverb_amount > 0:
```

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,254 @@ def requires_actor(self):
...
+            props.modulationTime = self.modulation_time
+            props.modulationDepth = self.modulation_depth
+            props.airAbsorptionHF = self.air_absorption_hf
+            props.hfReference = self.hf_reference
+            props.lfReference = self.lf_reference
+            for flag in self.flags:
+                props.flags |= getattr(EAXReverbProperties, flag)
+            eax_listener.listenerProps = props
+        elif self.preset == "MORE":
+            eax_listener.listenerProps = getattr(EAXReverbProperties, "REVERB_PRESET_" + self.preset_more)
```
Hoikas, line 449:
```suggestion
            eax_listener.listenerProps = getattr(EAXReverbProperties, "REVERB_PRESET_{}".format(self.preset_more))
```

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,254 @@ def requires_actor(self):
...
+        else:
+            eax_listener.listenerProps = getattr(EAXReverbProperties, "REVERB_PRESET_" + self.preset)
```
Hoikas, line 451: see above

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,254 @@ def requires_actor(self):
...
+
+    # Room rolloff - always at 0 in all presets, so screw it.
+    # room_rolloff_factor = FloatProperty(name="Room Rolloff Factor", description="Room Rolloff Factor",
+                                        # default=0.0, min=0.0, max=1.0)
+
+    flags = EnumProperty(name="Flags",
+                         description="Reverb flags",
+                         items=reverb_flags,
+                         default={ "kFlagDecayTimeScale", "kFlagReflectionsScale", "kFlagReflectionsDelayScale",
+                                   "kFlagReverbScale", "kFlagReverbDelayScale", "kFlagEchoTimeScale" },
```
Hoikas, line 416:
```suggestion
                         default={"kFlagDecayTimeScale", "kFlagReflectionsScale", "kFlagReflectionsDelayScale",
                                   "kFlagReverbScale", "kFlagReverbDelayScale", "kFlagEchoTimeScale"},
```

korman/properties/modifiers/region.py
```diff
@@ -203,6 +203,254 @@ def requires_actor(self):
...
+    hf_reference = FloatProperty(name="HF reference", description="High Frequency Reference",
+                                 default=5000.0, min=1000.0, max=20000.0,
+                                 options=set())
+    lf_reference = FloatProperty(name="LF reference", description="Low Frequency Reference",
+                                 default=250.0, min=20.0, max=1000.0,
+                                 options=set())
+
+    # Room rolloff - always at 0 in all presets, so screw it.
+    # room_rolloff_factor = FloatProperty(name="Room Rolloff Factor", description="Room Rolloff Factor",
+                                        # default=0.0, min=0.0, max=1.0)
```
Hoikas, line 410: Maybe just leave this out?

Jrius:
Thanks for the review, I implemented all suggestions. Let me know if you see something else.

Hoikas:
This LGTM, but I think I would like to wait for client support in H-uru/Plasma before merging (this is you cue, @Deledrius 😄).

patmauro:
Hey folks - found this PR, and unless @Jrius has any other adjustments in mind, I'd like to advocate for re-assessing the hold that was placed on this in 2021, and considering merging this proactively.
1. While H-uru/Plasma#1000 is not _techincally_ complete, as of Feb 20 there is now, for the first time, a working client build that supports EFX, even if it is still just a draft. Recently, Jrius also included EAX effects in his RAD contest Age, using a Korman branch including this feature, and we were able to successfully test and reproduce the effects in a client built from Deledrius' draft PR. Given the circumstances, I think this might be a very good time to give Age builders the ability to start laying the groundwork needed to support this feature once it officially goes live - this will help avoid there being an awkward period of time during which this seems to be only _half_-supported due to many Ages lacking the feature.
2. Importantly, we noticed that lack of EAX support in Korman has already unfortunately resulted in several Cyan Ages having their EAX data/info destroyed as a result of Korman conversion, impacting at least **Watcher's Pub**, **K'veer**, and **Kirel**, if not more. As a preventative measure, making sure this feature is available would likely help avoid any future loss, even if clients do not fully/officially support this yet.

While I totally understand the hesitation to deploying this until the client officially supports it, I really think there's no downside to providing folks the tools needed to plan for this in advance and prevent a "mad scramble" to implement this down the line - especially since providing this option will also help prevent feature _loss_ in any Cyan content. I really think it's time to consider merging this one.

Deledrius:
I agree with your assessment, especially regarding Age conversions.  Now that it's possible to test this PR (even if preliminarily), it makes sense to be proactive in the tool support, to me.

Hoikas (approved):

PR #254, opened 2021-06-09
Support for quaternion rotation animations

Hazado:
Put this together, through trial and error
There are probably better way to do this
Was able to export quaternion animation with it

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -388,10 +388,9 @@ class AnimationConverter:
         if not fcurves and not allow_empty:
             return None
         
-        rotation_quaternion = (i for i in fcurves if i.data_path == "rotation_quaternion")
+        rotation_quaternion = any((i.data_path == "rotation_quaternion" for i in fcurves))
         
         pos = self.make_pos_controller(fcurves, "location", xform.to_translation())
-        # TODO: support rotation_quaternion
         if rotation_quaternion:
             rot = self.make_rot_controller(fcurves, "rotation_quaternion", xform.to_quaternion(), 4)
         else:
@@ -475,7 +474,7 @@ class AnimationConverter:
         if bez_chans:
             ctrl = self._make_scalar_compound_controller(keyframes, bez_chans)
         else:
-            ctrl = self._make_quat_controller( keyframes, num_channels)
+            ctrl = self._make_quat_controller(keyframes, num_channels)
         return ctrl
 
     def make_scale_controller(self, fcurves, data_path : str, default_xform, convert=None) -> plLeafController:
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -388,9 +388,9 @@ class AnimationConverter:
         if not fcurves and not allow_empty:
             return None
 
-        pos = self.make_pos_controller(fcurves, xform.to_translation())
+        pos = self.make_pos_controller(fcurves, "location", xform.to_translation())
         rot = self.make_rot_controller(fcurves, rotation_mode, xform)
-        scale = self.make_scale_controller(fcurves, xform.to_scale())
+        scale = self.make_scale_controller(fcurves, "scale", xform.to_scale())
         if pos is None and rot is None and scale is None:
             if not allow_empty:
                 return None
@@ -447,8 +447,8 @@ class AnimationConverter:
         # Now we make the controller
         return self._make_matrix44_controller(keyframes)
 
-    def make_pos_controller(self, fcurves, default_xform, convert=None) -> Union[None, plLeafController]:
-        pos_curves = [i for i in fcurves if i.data_path == "location" and i.keyframe_points]
+    def make_pos_controller(self, fcurves, data_path: str, default_xform, convert=None) -> Union[None, plLeafController]:
+        pos_curves = [i for i in fcurves if i.data_path == data_path and i.keyframe_points]
         keyframes, bez_chans = self._process_keyframes(pos_curves, 3, default_xform, convert)
         if not keyframes:
             return None
@@ -507,8 +507,8 @@ class AnimationConverter:
                 else:
                     return self._make_quat_controller(keyframes)
 
-    def make_scale_controller(self, fcurves, default_xform, convert=None) -> plLeafController:
-        scale_curves = [i for i in fcurves if i.data_path == "scale" and i.keyframe_points]
+    def make_scale_controller(self, fcurves, data_path: str, default_xform, convert=None) -> plLeafController:
+        scale_curves = [i for i in fcurves if i.data_path == data_path and i.keyframe_points]
         keyframes, bez_chans = self._process_keyframes(scale_curves, 3, default_xform, convert)
         if not keyframes:
             return None
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -387,10 +387,15 @@ class AnimationConverter:
     def convert_transform_controller(self, fcurves, xform, allow_empty=False) -> Union[None, plCompoundController]:
         if not fcurves and not allow_empty:
             return None
-
+        
+        rotation_quaternion = (i for i in fcurves if i.data_path == "rotation_quaternion")
+        
         pos = self.make_pos_controller(fcurves, "location", xform.to_translation())
         # TODO: support rotation_quaternion
-        rot = self.make_rot_controller(fcurves, "rotation_euler", xform.to_euler())
+        if rotation_quaternion:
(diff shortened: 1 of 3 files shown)


DoobesURU:
I just tried this with Direbo (the butterflies' wings have this type of animation) and got an error pertaining to line 767 (`_process_keyframes`) of the animation.py script:

"ZeroDivisionError: float division by zero"

Jrius:
> 
> I just tried this with Direbo (the butterflies' wings have this type of animation)

This might also be a bug with ZLZ itself. It generates some invalid quaternion keyframes for the Relto butterflies animations (something about compressed quat keyframes that I haven't fixed yet).

korman/exporter/animation.py
```diff
@@ -387,10 +387,15 @@ def _convert_transform_animation(self, name, fcurves, xform, allow_empty=False)
     def convert_transform_controller(self, fcurves, xform, allow_empty=False) -> Union[None, plCompoundController]:
         if not fcurves and not allow_empty:
             return None
-
+        
+        rotation_quaternion = (i for i in fcurves if i.data_path == "rotation_quaternion")
```
Hoikas, line 391:
```suggestion
        rotation_quaternion = any((i.data_path == "rotation_quaternion" for i in fcurves))
```

korman/exporter/animation.py
```diff
@@ -387,10 +387,15 @@ def _convert_transform_animation(self, name, fcurves, xform, allow_empty=False)
...
+        
         pos = self.make_pos_controller(fcurves, "location", xform.to_translation())
         # TODO: support rotation_quaternion
```
Hoikas, line 394:
```suggestion
```

korman/exporter/animation.py
```diff
@@ -470,7 +475,7 @@ def make_rot_controller(self, fcurves, data_path : str, default_xform, convert=N
         if bez_chans:
             ctrl = self._make_scalar_compound_controller(keyframes, bez_chans)
         else:
-            ctrl = self._make_quat_controller( keyframes)
+            ctrl = self._make_quat_controller( keyframes, num_channels)
```
Hoikas, line 478:
```suggestion
            ctrl = self._make_quat_controller(keyframes, num_channels)
```

Hoikas (commented):
I think the reason I didn't support quaternions was that I wasn't sure if objects could have both quat and euler keyframes simultaneously. If so, we'd be in a bit of a pickle about which set of fcurves to use.

DoobesURU:
> 
> 
> > I just tried this with Direbo (the butterflies' wings have this type of animation)
> 
> This might also be a bug with ZLZ itself. It generates some invalid quaternion keyframes for the Relto butterflies animations (something about compressed quat keyframes that I haven't fixed yet).

Most likely. The good news is ZLZ can import Myst V Ages. The bad news is the animations are "squished"...ie they need to be multiplied by, as I've found, a factor of 60. Not sure if this is something that can be incorporated given we *may* be using some of that content in the future. 😉

Hazado:
> I think the reason I didn't support quaternions was that I wasn't sure if objects could have both quat and euler keyframes simultaneously. If so, we'd be in a bit of a pickle about which set of fcurves to use.

I believe they can in blender as they will both show up in the graphs window.
But I don't think it would work in game if they were both there.
If we cant seperate the fcurves, then known issue, pick either euler or quat for your animations. No mix and match.

Hazado:
Tested an animation that had both xyz euler and quaternion animation keyframes.
It always picked the quaternion animation keyframes if they were present.

I had XYZ Euler selected during the export.

Removing the Quat keyframes completely and it exported XYZ Euler again.

Hoikas:
Are the contents of the euler and quat keyframes identical or can they differ? If they can differ, our job becomes much more complex.

Hazado:
They can differ in blender. Each animation keyframe is separated from each other in blender.
The rotation of one will affect the other if you switch between them, but once keyframes are set they stay the same.

Quat Keyframe added, jump to euler, Euler coords calculated based on objects current orientation. Euler coords moved around, set at same time. Jump back to quat, quat coords calculated based on objects current orientation. move away and back, quat keyframes set earlier are back.

Hope the above explains my experiences.

Hoikas:
Ok, so what we will probably have to do is come up with a smarter converter than this... Uru can have bezier and linear euler keyframes but only linear quaternion keyframes. (So we'll probably only be able to export quaternions if all keyframes are set to linear.) We'll need our converter to be able to handle euler (3 channels), quaternion (4 channels), and axis angle (4 channels) keyframes from the fcurves... Then we can dump the output to Blender. Joy.

Jrius:
(Slightly off topic)

> 
> 
> Most likely. The good news is ZLZ can import Myst V Ages. The bad news is the animations are "squished"...ie they need to be multiplied by, as I've found, a factor of 60. Not sure if this is something that can be incorporated given we _may_ be using some of that content in the future. 😉

Noted, will fix soon. Animations have always been a weak point of ZLZ, unfortunately. But at least, as you've found, relying on HSPlasma means Ages from any game should import (mostly) well.

Hoikas:
@Hazado @DoobesURU I have just pushed a change to this pull request. Unlike my previous statements, it appears that @Hazado was correct in that we should export the keyframe type that's been selected in the Blender object. So, now you will export the rotation FCurves indicated by "Rotation Mode". Quaternions and axis angle rotations always go out as quaternions with linear interpolation. Eulers will only go out as quaternions if linear interpolation is selected, otherwise they are exported as eulers. I additionally fixed a nasty bug where only XYZ Eulers would work correctly in game. Give it a test run and let me know if there are any problems!

Hazado:
Tested the changes Hoikas made and they work great.

I specifically tested the following animation types
Axis Angle 		Linear
Quat 			Linear
zyx Eluer		Linear
xyz Euler		Bezier

All exported and worked in game how I expected them too.

DoobesURU:
Unfortunately, I got this error on an object with the texture's opacity animated:

```
Error: make_pos_controller() takes from 3 to 4 positional arguments but 5 were given
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\operators\op_export.py", line 220, in execute
    e.run()
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\convert.py", line 98, in run
    self._export_scene_objects()
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\convert.py", line 266, in _export_scene_objects
    export_fn(sceneobject, bl_obj)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\convert.py", line 290, in _export_mesh_blobj
    self.mesh.export_object(bo, so)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\mesh.py", line 585, in export_object
    drawables = self._export_mesh(bo)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\mesh.py", line 609, in _export_mesh
    geospans, mat2span_LUT = self._export_material_spans(bo, mesh, materials)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\mesh.py", line 653, in _export_material_spans
    matKey = self.material.export_material(bo, blmat)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\material.py", line 233, in export_material
    tex_layer = self.export_texture_slot(bo, bm, hsgmat, slot, idx, name=tex_name)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\material.py", line 500, in export_texture_slot
    layer = self._export_layer_animations(bo, bm, slot, idx, layer)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\material.py", line 539, in _export_layer_animations
    ctrl = converter(bo, bm, tex_slot, base_layer, fcurves)
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\material.py", line 588, in _export_layer_diffuse_animation
    ctrl = self._exporter().animation.make_pos_controller(fcurves, "diffuse_color", bm.diffuse_color, translate_color)
TypeError: make_pos_controller() takes from 3 to 4 positional arguments but 5 were given
```

For the record, it's the KI symbol glow on one of the imagers.

DoobesURU:
Commit 715ac3f fixed the above issue. Everything appears to be working now!

PR #256, opened 2021-06-12
Add support for waveset buoys

dpogue:
Adds a Buoy modifier that can be attached to a waveset. That modifier contains a list of objects that should float on top of the waveset.

korman-buoys

@Hoikas It seems to me like it would be best to have the list of buoys directly reference the objects, but everywhere else I see these sorts of lists in korman modifiers there's always a class in the middle (`PlasmaBuoyObject` in this case)

### DIFF
--- a/korman/properties/modifiers/water.py
+++ b/korman/properties/modifiers/water.py
@@ -471,3 +471,36 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
     def export(self, exporter, bo, so):
         waveset = exporter.mgr.find_create_object(plWaveSet7, name=bo.name, so=so)
         self.convert_wavestate(waveset.state.texState)
+
+
+class PlasmaBuoyObject(idprops.IDPropObjectMixin, bpy.types.PropertyGroup):
+    display_name = StringProperty(name="Display Name")
+    buoy_object = PointerProperty(name="Buoy Object",
+                                  description="Object that float on water",
+                                  type=bpy.types.Object,
+                                  poll=idprops.poll_mesh_objects)
+
+    @classmethod
+    def _idprop_mapping(cls):
+        return {"buoy_object": "object_name"}
+
+
+class PlasmaWaterBuoyModifier(PlasmaModifierProperties):
+    pl_depends = {"water_basic"}
+    pl_id = "water_buoy"
+
+    bl_category = "Water"
+    bl_label = "Water Buoys"
+    bl_description = ""
+
+    buoys = CollectionProperty(type=PlasmaBuoyObject)
+    active_buoy_index = IntProperty(options={"HIDDEN"})
+
+    def export(self, exporter, bo, so):
+        waveset = exporter.mgr.find_create_object(plWaveSet7, name=bo.name, so=so)
+        waveset.setFlag(plWaveSet7.kHasBuoys, True)
+
+        for i in self.buoys:
+            if i.buoy_object is None:
+                raise ExportError("'{}': Buoy Object for '{}' is invalid".format(self.key_name, i.display_name))
+            waveset.addBuoy(exporter.mgr.find_create_key(plSceneObject, bl=i.buoy_object))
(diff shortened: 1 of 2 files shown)


korman/properties/modifiers/water.py
```diff
@@ -471,3 +471,36 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
...
+class PlasmaBuoyObject(idprops.IDPropObjectMixin, bpy.types.PropertyGroup):
+    display_name = StringProperty(name="Display Name")
+    buoy_object = PointerProperty(name="Buoy Object",
+                                  description="Object that float on water",
+                                  type=bpy.types.Object,
+                                  poll=idprops.poll_mesh_objects)
+
+    @classmethod
+    def _idprop_mapping(cls):
+        return {"buoy_object": "object_name"}
```
Hoikas, line 485:
```suggestion
```

This is only needed for old string properties to be updated to pointer properties. No new property groups need it.

korman/properties/modifiers/water.py
```diff
@@ -471,3 +471,36 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
...
+    buoys = CollectionProperty(type=PlasmaBuoyObject)
+    active_buoy_index = IntProperty(options={"HIDDEN"})
+
+    def export(self, exporter, bo, so):
+        waveset = exporter.mgr.find_create_object(plWaveSet7, name=bo.name, so=so)
+        waveset.setFlag(plWaveSet7.kHasBuoys, True)
+
+        for i in self.buoys:
+            if i.buoy_object is None:
+                raise ExportError("'{}': Buoy Object for '{}' is invalid".format(self.key_name, i.display_name))
```
Hoikas, line 505:
```suggestion
                raise ExportError("'{}': Buoy Object for '{}' is invalid", self.key_name, i.display_name)
```

korman/properties/modifiers/water.py
```diff
@@ -471,3 +471,36 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
     def export(self, exporter, bo, so):
         waveset = exporter.mgr.find_create_object(plWaveSet7, name=bo.name, so=so)
         self.convert_wavestate(waveset.state.texState)
+
+
+class PlasmaBuoyObject(idprops.IDPropObjectMixin, bpy.types.PropertyGroup):
+    display_name = StringProperty(name="Display Name")
```
Hoikas, line 477: I've been migrating toward having these lists display the actual name of the object or "[Empty]" if unset. IMO, it makes things less tedious.

Hoikas:
Also, we should probably issue a porting message that buoys only function on MOUL.

dpogue:
> Also, we should probably issue a porting message that buoys only function on MOUL.

Thinking a bit more about edge cases, we probably also want something that checks that the buoy is a physical dynamic object (although the engine should ignore any non-physical buoys)

Hoikas:
I think this modifier should call the physics exporter - it's an idempotent process so no worries about stepping all over anything.

korman/properties/modifiers/water.py
```diff
@@ -471,3 +471,38 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
...
         waveset = exporter.mgr.find_create_object(plWaveSet7, name=bo.name, so=so)
         self.convert_wavestate(waveset.state.texState)
+
+
+class PlasmaBuoyObject(bpy.types.PropertyGroup):
+    buoy_object = PointerProperty(name="Buoy Object",
+                                  description="Object that float on water",
+                                  options=set(),
+                                  type=bpy.types.Object,
+                                  poll=idprops.poll_dynamic_objects)
```
Hoikas, line 481: While this is correct from an implementation standpoint, I feel like it's putting too much of a requirement on the user on knowing how buoys work under the hood. Instead, I think we should call the physics exporter in the export fxn and place the resulting physical in the dynamic group... `exporter.physics.generate_physical(bo, so, member_group="kGroupDynamic")` This poll fxn should probably just check for mesh objects. Further, we should alert in the UI (and maybe the log too?) if the collision mod is on with dynamic unticked. Alternatively, add instrumentation to the physics exporter to force buoys dynamic.

PR #257, opened 2021-06-26
Ensure VtxAlpha materials are alpha blended.

Hoikas:


Hoikas:
This never worked correctly and is out of date now. Closing.

PR #258, opened 2021-06-26
Alpha Vertex Blend Toggle

DoobesURU:
Adds a per-layer toggle just below the opacity slider to use the Alpha VCol values and blend the texture. Still needs sanity checks.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -509,7 +509,10 @@ class MaterialConverter:
         else:
             layer_props = texture.plasma_layer
             layer.opacity = layer_props.opacity / 100
-            self._handle_layer_opacity(layer, layer_props.opacity)
+            if layer_props.opacity < 100 and not state.blendFlags & hsGMatState.kBlendMask:
+                state.blendFlags |= hsGMatState.kBlendAlpha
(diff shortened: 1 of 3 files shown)


DoobesURU:
Just curious if we need anything else added/subtracted from this to get 'er out the door or if the approach is too troublesome.

Hoikas:
I wanted to spend some time trying to understand how all this works and if there's a better way to expose what's going in.

DoobesURU:
> I wanted to spend some time trying to understand how all this works and if there's a better way to expose what's going in.

Ah, I understand. Like I had said before, the AlphaBlends are an odd bit of business.

DoobesURU:
Just to revive this discussion, perhaps it'd be better to use the method that (_eep_) PyPRP utilized, ie when there's an "Alpha" vertex color, it's used on every layer after the first? That way, it's not quite as exposed as a direct toggle, but the downside would be the end user would lose a bit of control as to which textures would receive the alpha.

Unless maybe we tie it to the existing `texture_slot.use_map_alpha` toggle? And anything with that on plus the Alpha VCol would get the blending?

PR #259, opened 2021-06-29
Fix Offset Y

DoobesURU:
Small change to fix the Y value of the layer offset from the previous PR #243 , which was causing it to export in reverse.

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -427,7 +427,7 @@ class AnimationConverter:
             scale = kwargs[scale_path]
 
             translation = hsVector3(pos[0] - (scale[0] - 1.0) / 2.0,
-                                    -pos[1] - (scale[1] - 1.0) / 2.0,
+                                    pos[1] - (scale[1] - 1.0) / 2.0,
                                     pos[2] - (scale[2] - 1.0) / 2.0)
             matrix = hsMatrix44()
             matrix.setTranslate(translation)


Jrius:
Are you sure ? The whole point of #243 was to mimics Blender's behavior :
output
The reason behind that negation is to convert Blender's UV space from OpenGL to DirectX. The tradeoff I mentioned is that it _does_ break animations in existing Ages, which IMHO is a necessary evil to get consistent behavior across the two programs (fixing it simply requires mirroring the afflicted anims on the Y axis in the IPO curve editor).

DoobesURU:
While this was an existing Age (actually a ZLZ import of an Age), the offset animation was freshly redone with the most up-to-date scripts. Are you sure this wasn't just a typo in the code?

Jrius:
I pulled and retested again on PotS. Current (Y negation) gives an animation that looks like Blender's, your proposal (everything positive) makes the texture slide in the wrong direction.
Does the animation look good when rendering it in Blender ? Does the layer or the object have any kind of scaling ?

DoobesURU:
This may be another one of those "good for PotS, bad for MOULa" kind of things that tend to crop up. All I know is, when I export for MOULa, texture offsets on the Y axis are going the wrong way. Removing the "-" seems to fix it, at least for MOUL export. 🤷‍♂️

Jrius:
> I don't generally render anything in Blender

Could it be you just rebuilt the texture anims by copying values from PrpShop's plLayerAnims, and never preview how Blender itself plays those animations before exporting ? That's the problem then.

Blender does enforce a particular scroll direction along the Y UV axis, which is NOT the same as Plasma's. However, Blender's viewport sucks and cannot playback texture animations in real time. So the easiest way to preview an anim is usually to use full renders. (For simplicity, put a camera in front of your object, turn off shadowing/raytracing to make rendering faster, hit Ctrl+F12, wait for enough frames to be rendered, hit escape, then Ctrl+F11 - that will play back the rendered frames in real time.) Please try doing that with your animated texture, and see if the texture scroll direction is what you would expect to see in Plasma.

Like I said, Blender's UV Y axis is the inverse from Plasma (long story short, this is a neverending war between OpenGL and DirectX applications). Up until now, Korman didn't take this issue into account for animations, so when you exported, texture anims were flipped in Plasma. #243 corrected this so that Blender behavior == Plasma behavior. BUT, if you're rebuilding texture anims from PRP values, then you have to always remember to negate Y offset values, because that's what Blender itself expects. Korman will flip those values back on export.

Hoikas:
Based on discussions and [photographic evidence](https://www.youtube.com/watch?v=CRTNLLdgI_0), @Jrius's approach is correct. Korman was simply wrong from the beginning of time(TM).

PR #261, opened 2021-07-28
Persistent Baked Lighting

Hoikas:
This changeset allows reusing the same baked lighting from one export to another. This is an opt-in feature for when your Age becomes lighting heavy and the performance penalty becomes expensive when trying to iterate on other aspects of the Age.

New operators have been added. In the toolbox, you can "Bake All" or "Bake Selected" lighting at any time. In the lightmap modifier, you can "Bake for Export". These buttons bake vertex colors to the "Col" layer or set the new "Image" field on the lightmap modifier, as appropriate. On export, if a "Col" or image texture is found, it is used instead of lighting being rebaked. This means that if the lighting changes on those objects YOU have to rebake them using these same features. To return to the previous functionality of rebake every export, the toolbox operators "Clear All" or "Clear Selection" may be used. Alternatively, remove the image from the lightmap modifier or delete the Col vertex color layer.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -56,6 +56,7 @@ class Exporter:
             self.image = image.ImageCache(self)
             self.locman = locman.LocalizationConverter(self)
             self.decal = decal.DecalConverter(self)
+            self.oven = etlight.LightBaker(self.report)
 
             # Step 0.8: Init the progress mgr
             self.mesh.add_progress_presteps(self.report)
@@ -126,10 +127,8 @@ class Exporter:
                 self.report.raise_errors()
 
     def _bake_static_lighting(self):
-        lighting_method = self._op.lighting_method
-        if lighting_method != "skip":
-            oven = etlight.LightBaker(self.report)
-            oven.bake_static_lighting(self._objects)
+        if self._op.lighting_method != "skip":
+            self.oven.bake_static_lighting(self._objects)
 
     def _collect_objects(self):
         scene = bpy.context.scene
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -14,10 +14,13 @@
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
 import bpy
-from ..korlib import ConsoleToggler
+
 from pathlib import Path
+from contextlib import ExitStack
+
+from ..korlib import ConsoleToggler
+
 from PyHSPlasma import *
-import time
 
 from . import animation
 from . import camera
@@ -44,7 +47,7 @@ class Exporter:
 
     def run(self):
         log = logger.ExportVerboseLogger if self._op.verbose else logger.ExportProgressLogger
-        with ConsoleToggler(self._op.show_console), log(self._op.filepath) as self.report:
+        with ConsoleToggler(self._op.show_console), log(self._op.filepath) as self.report, ExitStack() as self.context_stack:
             # Step 0: Init export resmgr and stuff
             self.mgr = manager.ExportManager(self)
             self.mesh = mesh.MeshConverter(self)
@@ -56,7 +59,7 @@ class Exporter:
             self.image = image.ImageCache(self)
             self.locman = locman.LocalizationConverter(self)
             self.decal = decal.DecalConverter(self)
-            self.oven = etlight.LightBaker(self.report)
+            self.oven = etlight.LightBaker(self.report, stack=self.context_stack)
 
             # Step 0.8: Init the progress mgr
             self.mesh.add_progress_presteps(self.report)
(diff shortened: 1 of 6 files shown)


Hoikas:
cc @DoobesURU

DoobesURU:
I did a (very early in the morning) test run with Chiso. Quite a few items came out a bit dark, but it might be due to lingering Col lighting that simply doesn't match.

I'll give it another try when I get home later and report back.

DoobesURU:
I'm going to do a couple more exports today and report back on the results.

DoobesURU:
Note that previously, before all those recent commits, that the Bake All function seemed to fudge some of the vertex lighting (in this case, Chiso Preniv's). Clear/Bake Selection and the Bake for Export buttons all seem to bake things correctly.

When I get home from work today, I'll try another Clear All/Bake All to see what happens with the latest fixes.

Jrius:
I wanted to try it out to give you some extra feedback, but the "Bake for Export" button generates the following error on my end.

```Traceback (most recent call last):
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\operators\op_lightmap.py", line 100, in invoke
    return self.execute(context)
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\operators\op_lightmap.py", line 66, in execute
    with self._oven(context) as bake:
  File "C:\Program Files\Blender Foundation\Blender\2.79\python\lib\contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\operators\op_lightmap.py", line 37, in _oven
    with UiHelper(context), ConsoleToggler(console), LightBaker(verbose=verbose) as oven:
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\exporter\etlight.py", line 35, in __init__
    self.add_progress_steps(self._report, True)
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\exporter\etlight.py", line 57, in add_progress_steps
    _MeshManager.add_progress_presteps(report)
  File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\korman\exporter\mesh.py", line 172, in add_progress_presteps
    report.progress_add_step("Applying Blender Mods")
TypeError: progress_add_step() missing 1 required positional argument: 'name'
```

Either way, it's cool to read that Korman no longer fiddles with UV mapping - that's been an issue I ran into and had to work around. Does it still generate a valid UV map if the artist doesn't supply one in the lightmap modifier ? Using the diffuse UV map is generally incorrect because it often contains overlapping UVs (for artistic purposes), which screw up lightmapping.

Jrius:
About smoothing vertex colors - this is not a good idea, or at least there is a much better way to do it.

Vertex colors should not be blindly smoothed. Different vcol per-loop is intentional, because artists often mix smooth and sharp edges to hint Blender into splitting/merging vertex normals per-loop (thus altering the resulting lighting). It's the equivalent to 3dsMax's smoothing groups. ZLZ also figures out edge sharpness when importing for the same purpose - it's a really useful feature. So smoothing vcol blindly will render sharp edges useless while baking to vcol, which is undesirable.

I suppose what you're attempting to do by smoothing vertex colors is to work around this vertex "[self-occluding](https://blenderartists.org/t/vertex-color-baking-artifacts/653442)" bug (for lack of a better word). I have it all over my Canyon Age, and it seems to be caused by soft shadows and AO.

I've got a better solution on one of my branches, which smooths vertex color only if they share a smooth edge. It also takes the brightest color instead of averaging them, which completely discards the bogus black color (much better). The results are near-perfect from what I can tell.
I'll update and submit it once this PR is merged.

Hoikas:
@DoobesURU and I have been iterating on this quite a bit, and some of the commentary here and in the commits got a bit out of date... After experimenting, I found that I would get unexpected results after Korman's UV processing was removed. What problems were you having to work around with that? I may try pulling it out again at some point. I have fixed the traceback, though!

> I'll update and submit it once this PR is merged.

That sounds great, so I have pulled out the vertex color smoothing commit from the PR.

DoobesURU:
> What problems were you having to work around with that?

IIRC, I thought there were seams, but it turns out that a) the wall I was looking at looked that way even before the new PR and b) the other seams I found were due to the vertex color not being smoothed properly.

So I think as far as the processing, you could probably go ahead and remove it. I'll give things another go if/when you do.

DoobesURU:
Looks like the latest fixes have done the trick. I have baked at export, Baked All, and also saved, exited and re-exported. All of it is working quite well now, AFAICT.

Hoikas:
:shipit:

Deledrius (approved):

PR #269, opened 2021-08-03
Don't touch my UV maps!

### LINKED ISSUE #265: Don't average/pack lightmap UVs.
Currently, Korman averages and packs UV islands in an artist specified UV texture for lightmaps. This is undesirable in light of #261, which will enable the lightmap texture to persist. Additionally, @Jrius alludes to other [issues](https://github.com/H-uru/korman/pull/261#issuecomment-890504917) with this forced process, however there was no elaboration about what those issues were. However, it is a requirement in the case of objects with Blender modifiers - without this process, lighting will be duplicated onto newly generated sections of mesh.

There are a few ways we can proceed:
- Only do the average/pack process if we detect modifiers on the object.
   - This may result in surprising behavior?
- Add a checkbox on the lightmap modifier for enabling the forced average/pack process.
   - What should the default value be?
   - This seems overly tedious.
- Deprecate artist supplied UV maps on objects with modifiers. Conditionally allow it with the average/pack process if an certain export policy is set, otherwise, error.
    - Export policies are currently in discussion on IRC. They would look and work somewhat like [CMake Policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html).

Hoikas:
It's better to trust the artist to know what they are doing if they provide a uv map for the lightmap modifier to use. Closes #265.

### DIFF
--- a/korman/exporter/etlight.py
+++ b/korman/exporter/etlight.py
@@ -421,6 +421,7 @@ class LightBaker:
                                   bo.name, modifier.uv_map, indent=2)
                 with self._set_mode("EDIT"):
                     bpy.ops.mesh.select_all(action="SELECT")
+                    bpy.ops.uv.select_all(action="SELECT")
                     bpy.ops.uv.pack_islands(margin=0.01)
         else:
             # same thread, see Sirius's suggestion RE smart unwrap. this seems to yield good
@@ -430,7 +431,6 @@ class LightBaker:
             self._associate_image_with_uvtex(uvtex, image)
             with self._set_mode("EDIT"):
                 bpy.ops.mesh.select_all(action="SELECT")
-                bpy.ops.uv.select_all(action="SELECT")
                 bpy.ops.uv.smart_project(island_margin=0.05)
 
     def _prep_for_vcols(self, bo, toggle):
--- a/korman/exporter/etlight.py
+++ b/korman/exporter/etlight.py
@@ -110,11 +110,11 @@ class LightBaker:
         """Bakes all static lighting for Plasma geometry"""
 
         self._report.msg("\nBaking Static Lighting...")
-        bake = self._harvest_bakable_objects(objs)
 
         with GoodNeighbor() as toggle:
             try:
                 # reduce the amount of indentation
+                bake = self._harvest_bakable_objects(objs, toggle)
                 result = self._bake_static_lighting(bake, toggle)
             finally:
                 # this stuff has been observed to be problematic with GoodNeighbor
@@ -238,7 +238,7 @@ class LightBaker:
                 return True
         return False
 
-    def _harvest_bakable_objects(self, objs):
+    def _harvest_bakable_objects(self, objs, toggle):
         # The goal here is to minimize the calls to bake_image, so we are going to collect everything
         # that needs to be baked and sort it out by configuration.
         default_layers = tuple((True,) * _NUM_RENDER_LAYERS)
@@ -255,7 +255,8 @@ class LightBaker:
                     if self.lightmap_uvtex_name in uv_texture_names:
                         self._report.msg("'{}': Skipping due to valid lightmap override", obj.name, indent=1)
                     else:
-                        self._report.msg("'{}': Have lightmap but UVs are missing???", obj.name, indent=1)
+                        self._report.warn("'{}': Have lightmap, but regenerating UVs", obj.name, indent=1)
+                        self._prep_for_lightmap_uvs(obj, mod.image, toggle)
                     return False
                 return True
             return False
@@ -360,6 +361,24 @@ class LightBaker:
             im = data_images.new(im_name, width=size, height=size)
         self._lightmap_images[bo.name] = im
 
+        self._prep_for_lightmap_uvs(bo, im, toggle)
+
+        # Now, set the new LIGHTMAPGEN uv layer as what we want to render to...
+        # NOTE that this will need to be reset by us to what the user had previously
+        # Not using toggle.track due to observed oddities
+        for i in uv_textures:
+            value = i.name == self.lightmap_uvtex_name
+            i.active = value
+            i.active_render = value
+
+        # Indicate we should bake
+        return True
+
+    def _prep_for_lightmap_uvs(self, bo, image, toggle):
+        mesh = bo.data
+        modifier = bo.plasma_modifiers.lightmap
+        uv_textures = mesh.uv_textures
+
         # If there is a cached LIGHTMAPGEN uvtexture, nuke it
         uvtex = uv_textures.get(self.lightmap_uvtex_name, None)
         if uvtex is not None:
@@ -392,7 +411,7 @@ class LightBaker:
             # if the artist hid any UVs, they will not be baked to... fix this now
             with self._set_mode("EDIT"):
                 bpy.ops.uv.reveal()
-            self._associate_image_with_uvtex(uv_textures.active, im)
+            self._associate_image_with_uvtex(uv_textures.active, image)
 
             # Meshes with modifiers need to have islands packed to prevent generated vertices
             # from sharing UVs. Sigh.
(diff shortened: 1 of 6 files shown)


Deledrius (approved):

korman/exporter/etlight.py
```diff
@@ -375,6 +375,8 @@ def _prep_for_lightmap(self, bo, toggle):
         # this tended to create sharp edges. There was already a discussion about this on the
```
Jrius, line 375: (Commenting here, because Github won't let me comment further up)

Small fix in case the user wants to lightmap a textureless object:
```
# Because the way Blender tracks active UV layers is massively stupid...
if uv_textures.active:
    self._uvtexs[mesh.name] = uv_textures.active.name
```

korman/exporter/etlight.py
```diff
@@ -375,6 +375,8 @@ def _prep_for_lightmap(self, bo, toggle):
...
         # Guild of Writers forum, so I'm implementing a code version of dendwaler's process,
         # as detailed here: https://forum.guildofwriters.org/viewtopic.php?p=62572#p62572
+        # This has been amended with Sirius's observations in GH-265 about forced uv map
+        # packing. Namely, don't do it unless modifiers make us.
         uv_base = self._get_lightmap_uvtex(mesh, modifier)
```
Jrius, line 380: `_get_lightmap_uvtex` appears to still use any available UV channel for the lightmap. Unless the user specifies the UV to use in the lightmap modifier, this method should return nothing to let Korman generate the UVs by itself:
```
def _get_lightmap_uvtex(self, mesh, modifier):
    if modifier.uv_map:
        return mesh.uv_textures[modifier.uv_map]
    return None
```

korman/exporter/etlight.py
```diff
@@ -384,25 +386,25 @@ def _prep_for_lightmap(self, bo, toggle):
...
             self._associate_image_with_uvtex(uv_textures.active, im)
-            bpy.ops.object.mode_set(mode="EDIT")
 
-            # prep the uvtex for lightmapping
-            bpy.ops.mesh.select_all(action="SELECT")
-            bpy.ops.uv.average_islands_scale()
-            bpy.ops.uv.pack_islands()
+            # Meshes with modifiers need to have islands packed to prevent generated vertices
+            # from sharing UVs. Sigh.
+            if self.is_collapsed(bo):
```
Jrius, line 395: This never returns true. Looks like `LightBaker.__enter__` is never called due to lack of `with` statement. `MeshConverter.__enter__` is, though.

korman/exporter/etlight.py
```diff
@@ -384,25 +386,25 @@ def _prep_for_lightmap(self, bo, toggle):
...
+                self._report.warn("'{}': packing islands in UV Texture '{}' due to modifier collapse", bo.name, uv_base.name)
+                with self._set_mode("EDIT"):
+                    bpy.ops.mesh.select_all(action="SELECT")
+                    bpy.ops.uv.pack_islands()
```
Jrius, line 399:
```suggestion
                    bpy.ops.uv.pack_islands(margin=0.01)
```
The default of 0.001 packs islands a bit too tight IMHO. 0.01 should help avoid pixels leaking between islands, especially when dealing with many islands or small lightmap sizes.

korman/exporter/etlight.py
```diff
@@ -384,25 +386,25 @@ def _prep_for_lightmap(self, bo, toggle):
...
         else:
             # same thread, see Sirius's suggestion RE smart unwrap. this seems to yield good
             # results in my tests. it will be good enough for quick exports.
             uvtex = uv_textures.new(self.lightmap_uvtex_name)
```
Jrius, line 403:
```suggestion
            uvtex = uv_textures.new(self.lightmap_uvtex_name)
            uv_textures.active = uvtex
```
So apparently we never caught on to this small oversight with horrible consequences :rofl: We need to set the UVtex as active, otherwise Bad Things happen (like messing up the user's UVs, baking useless lightmaps, and increasing baking times tenfold).

korman/ui/modifiers/render.py
```diff
@@ -196,6 +196,8 @@ def lightmap(modifier, layout, context):
     col = layout.column()
     col.active = is_texture
     col.prop_search(modifier, "uv_map", context.active_object.data, "uv_textures")
+    if bool(modifier.id_data.modifiers):
```
Jrius, line 199:
```suggestion
    if bool(modifier.id_data.modifiers and modifier.uv_map):
```
No need to display the warning if the user specifies no UVmap, as we'll generate it ourselves.

Jrius (commented):
Works pretty well with the modifications I suggested. I went as far as remove all custom lightmap UVs in Tsoidahl Prad and let Korman generate those themselves, still looks perfect.
The only remaining issue is with `is_collapsed` not working due to the uninitialized `_MeshManager`. Meaning objects with modifiers and custom lightmap UVs aren't repacked correctly.

korman/exporter/etlight.py
Hoikas: OK, this is a bit of a chicken and egg problem because both `MeshConverter` and `LightBaker` are `_MeshManager`s. During the export process, this will always return False but it should work using the bake operators. I'll figure out what to do on that.

korman/ui/modifiers/render.py
Hoikas: This suggestion won't work as written, blender sequences do not implicitly convert to bool (bug in blender). So it needs to be:
```suggestion
    if bool(modifier.id_data.modifiers) and modifier.uv_map:
```

Jrius: Oops, right. Probably a copy paste error.

Jrius (changes requested):
UVs need to be selected before packing, but otherwise this works correctly !

We're hitting another problem though. The `LIGHTMAPGEN` uv is not persistent on meshes with modifiers, so if it's baked manually before export, export will fail with the message `Lightmap UV Texture 'LIGHTMAPGEN' seems to be missing. Did you delete it?`. So the object somehow needs to be re-run through `_prep_for_lightmap` (watch out in case lightmap size was changed), but without triggering `_bake_lightmaps`.

korman/exporter/etlight.py
```diff
@@ -384,25 +390,28 @@ def _prep_for_lightmap(self, bo, toggle):
...
+            uv_textures.active = uvtex
             self._associate_image_with_uvtex(uvtex, im)
-            bpy.ops.object.mode_set(mode="EDIT")
-            bpy.ops.mesh.select_all(action="SELECT")
-            bpy.ops.uv.smart_project()
-        bpy.ops.object.mode_set(mode="OBJECT")
+            with self._set_mode("EDIT"):
+                bpy.ops.mesh.select_all(action="SELECT")
```
Jrius, line 413:
```suggestion
                bpy.ops.mesh.select_all(action="SELECT")
                bpy.ops.uv.select_all(action="SELECT")
```
Otherwise island packing won't work.

Jrius (changes requested):
(I forgot to click submit on the line affected. D'oh !)

korman/exporter/etlight.py
Hoikas: Will this change also need to be replicated between lines 404 and 405?

Jrius: Nice catch - I actually clicked the wrong line when reviewing the second time. Yes, `bpy.ops.uv.select_all(action="SELECT")` must be placed before `bpy.ops.uv.pack_islands`, NOT before `bpy.ops.uv.smart_project` (the former operates on selected UV vertices, the latter on selected mesh vertices). Sorry for the confusion.

Jrius (approved):
Watch out for my mistake around `bpy.ops.uv.select_all` which needs to be moved a few lines up, but otherwise everything works great on my end. I can export a full lightmapped Age in 20 seconds, so far it works for all four types of objects :+1:

PR #270, opened 2021-08-05
Vcol artifacts removal

Jrius:
Based on Hoikas' PR, so marking this as draft for now.

Improved algorithm to remove vcol baking artifacts while preserving sharp edges. Most efficient on meshes mixing smooth and sharp edges. Some objects with lots of sharp edges might still display the dark color bugs - in my experience this rarely happens, more testing will probably be required due to the random nature of the bug. On an Age the size of Tsoidahl Prad, this process takes about 2.3 seconds.

Hats off if you understand the smoothing algorithm on the first read - the whole process of comparing vertices and loops is kinda tricky. Note that I'm also ramming the object through BMesh, so I had to manually track active vertex color layers instead of relying on GoodNeighbor.

### DIFF
--- a/korman/exporter/etlight.py
+++ b/korman/exporter/etlight.py
@@ -201,54 +201,47 @@ class LightBaker:
         # (If the object has an edge split modifier, well, screw you!)
         for bo in blender_objects:
             mesh = bo.data
-            bm = bmesh.new()
-            bm.from_mesh(mesh)
-
-            light_vcol = bm.loops.layers.color.get(self.vcol_layer_name)
-
-            # If no vertex color is found, then baking either failed (error raised by oven)
-            # or is turned off. Either way, bail out.
-            if light_vcol is None:
-                bm.free()
-                del bm
+            if self.vcol_layer_name not in mesh.vertex_colors:
+                # No vertex color. Baking either failed or is turned off.
                 continue
 
-            bm.faces.ensure_lookup_table()
+            with bmesh_from_object(bo) as bm:
+                bm.faces.ensure_lookup_table()
+                light_vcol = bm.loops.layers.color.get(self.vcol_layer_name)
 
-            for face in bm.faces:
-                for loop in face.loops:
-                    vert = loop.vert
-                    max_color = loop[light_vcol]
-                    if not face.smooth:
-                        # Face is sharp, so we can't smooth anything.
-                        continue
-                    # Now that we have a loop and its vertex, find all edges the vertex connects to.
-                    for edge in vert.link_edges:
-                        if len(edge.link_faces) != 2:
-                            # Either a border edge, or an abomination.
+                for face in bm.faces:
+                    for loop in face.loops:
+                        vert = loop.vert
+                        max_color = loop[light_vcol]
+                        if not face.smooth:
+                            # Face is sharp, so we can't smooth anything.
                             continue
-                        if not edge.smooth or (mesh.use_auto_smooth and
-                                               edge.calc_face_angle() > mesh.auto_smooth_angle):
-                            # Sharp edge. Don't care.
-                            continue
-                        if face in edge.link_faces:
-                            # Alright, this edge is connected to our loop AND our face.
-                            # Now for the Fun Stuff(c)... First, actually get ahold of the other
-                            # face (the one we're connected to via this edge).
-                            other_face = next(f for f in edge.link_faces if f != face)
-                            # Now get ahold of the loop sharing our vertex on the OTHER SIDE
-                            # of that damnable edge...
-                            other_loop = next(loop for loop in other_face.loops if loop.vert == vert)
-                            other_color = other_loop[light_vcol]
-                            # Phew ! Good, now just pick whichever color has the highest average value
-                            if sum(max_color) / 3 < sum(other_color) / 3:
-                                max_color = other_color
-                    # Assign our hard-earned color back
-                    loop[light_vcol] = max_color
+                        # Now that we have a loop and its vertex, find all edges the vertex connects to.
+                        for edge in vert.link_edges:
+                            if len(edge.link_faces) != 2:
+                                # Either a border edge, or an abomination.
+                                continue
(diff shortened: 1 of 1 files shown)


korman/exporter/etlight.py
```diff
@@ -179,13 +181,75 @@ def _bake_static_lighting(self, bake, toggle):
...
+        # For each loop, we find all other loops in the mesh sharing the same vertex, which aren't
+        # separated by a sharp edge. We then take the brightest color out of all those loops, and
+        # assign it back to the base loop.
+        # "Sharp edges" include edges manually tagged as sharp by the user, or part of a non-smooth
+        # face, or edges for which the face angle is superior to the mesh's auto-smooth threshold.
+        # (If the object has an edge split modifier, well, screw you!)
+        for bo in blender_objects:
+            mesh = bo.data
+            bm = bmesh.new()
+            bm.from_mesh(mesh)
```
Hoikas, line 205: I think I would feel better if this were a context manager to avoid the tedious and un-Pythonic manual resource management. So, you'd want something like a helper function:

```python
@contextmanager
def bmesh_edit_mesh(bl):
    bm = bmesh.new()
    bm.from_mesh(bl.data)
    try:
        yield bm
    except:
         pass
    else:
        bm.to_mesh(bl.data)
    finally:
        bm.free()
```

I'd probably use an exception to bail out of this context manager to avoid triggering the `bm.to_mesh` call, if absolutely needed - eg for line 211 `if light_vcol is None`.

Jrius: Right. Even better, with some tweaks I can simply reuse the existing `bmesh_from_object` from `helpers.py`, and avoid adding a new context manager.
Unfortunately, this also causes Blender to crash on `mesh.from_object(bl, bpy.context.scene)` the second time `bmesh_from_object` is called - for instance, when exporting a collision modifier. Seems specific to `from_object`, because `from_mesh` worked correctly earlier.
I don't have the means to debug the internals of Blender right now. If you have the time, [here is an example file](https://www.dropbox.com/s/jpev9ms2em8gvjq/TsoidahlPrad_279_kdebug_verylight.blend?dl=1) that causes the issue. Otherwise I'll see when I can setup a build environment to debug this.

Hoikas: It wouldn't surprise me if some of the shenanigans that we're doing with collapsing the modifiers and using a temporary mesh are causing problems with `from_object`. I think it might be best to just go ahead and add a new helper with `from_mesh()`. It's painfully easy to crash Blender from Python, so I I'd prefer to stick with what we know works 😄

Jrius: Understood. Debugging crashes on the Python side already takes a lot of time, so to be honest I really wasn't looking forward to digging into Blender itself just for the sake of it :laughing:

Hoikas: FWIW, I do have many working from source Blender builds for debugging, I just prefer to avoid actually doing that work. Especially when we have something we know works 😜

Jrius:
There we go. `_bmesh_from_mesh` is in the same class, let me know if you prefer it to be moved to the helpers file.
Should be ready for testing. @DoobesURU, due to the random nature of the bug I wouldn't mind having your opinion on how well this performs in your Ages, when you get the time to try it out.

DoobesURU:
For my most recent export of Descent, I got this series of errors:

```
Exception ignored in: <bound method LightBaker.__del__ of <korman.exporter.etlight.LightBaker object at 0x000001EA15522DA0>>
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\etlight.py", line 57, in __del__
    if self._own_report:
AttributeError: 'LightBaker' object has no attribute '_own_report'

Aborted after 0.00s
Error: __init__() got an unexpected keyword argument 'stack'
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\operators\op_export.py", line 220, in execute
    e.run()
  File "E:\Program Files (x86)\Blender Foundation\Blender 2.79\2.79\scripts\addons\korman\exporter\convert.py", line 62, in run
    self.oven = etlight.LightBaker(self.report, stack=self.context_stack)
TypeError: __init__() got an unexpected keyword argument 'stack'

location: <unknown location>:-1

location: <unknown location>:-1
```

DoobesURU:
Got a bit of weirdness when testing with Tiam:

image

Jrius:
Hm, I see... This object appears to be lightmapped, right ? In that case, that would be an issue with the lightmap UVs, more likely from the previous PR. Try to clear the baked lightmap and retry. If that still happens, upload the file somewhere and I'll have a look.

DoobesURU:
> 
> 
> Try to clear the baked lightmap and retry.

Note that this was a lightmap that was baked at export, but it turns out it wasn't your PR that did it...just some overlapping UV mapping that didn't seem to crop up until now. 🙂 

Anyway, once I gave them proper Lightmap UV maps instead of the default UVMaps used for the textures, it fixed things. Given that and my other exports so far, I believe this PR is fit for merging, IMHO.

Hoikas:
> default UVMaps used for the textures

Does Tiam explicitly use the texturing UV Maps for lightmaps?

DoobesURU:
> 
> 
> > default UVMaps used for the textures
> 
> Does Tiam explicitly use the texturing UV Maps for lightmaps?

In most cases, yes...and it usually isn't a problem given the way ametist mapped things. In the instance of the doors, however, it caused the two sides of the door that were sharing the symbol to overlap and create the nasty in the above textures.

Hoikas:
Ok, that seems fine. Like @Jirus said in the other PR, lightmap UV textures need to be made so that no faces overlap. It's pretty much dumb luck that the texturing UV maps worked before. We'll have to be sure to communicate this difference in the Korman 0.12 release post, though.

Thanks for the PR, Sirius. I think this will be a huge benefit for the automatic vertex coloring lighting!

Issue #262, opened 2021-07-30
Localized dynamic text modifier.


Hoikas:
The engine includes an `xDynTextObj.py` script that can use used to print localized text onto a texture. It would be nice to expose this in a simple modifier such that the localized contents are exposed the same way journals are. This would be a huge boon for simple note GUIs as well (see #136).

PR #273, opened 2021-08-12
Update LogicWiz modifiers for deferred logic export.

Hoikas:
In a previous changeset, the Advanced Logic modifier was changed to defer exporting logic until the very end of the export process. This means that many nodes are designed with the assumption that all non-logic PRP objects are fully exported by the time they are exported. LogicWiz modifiers, however, violated this assumption by exporting during the main export() phase.

So, this adds a new `pre_export()` phase for all modifiers that lets them generate logic trees and even entirely new Plasma Objects safely. Further, old LogicWiz modifiers have been tweaked to not leak junked objects if the export fails in the middle of those modifiers.

The ability to defer the export of generated node trees is a requirement for #262.


Deledrius (approved):
To the extent that I understand it, LGTM

PR #274, opened 2021-08-14
Fix weird node warping.

Hoikas:
Fixes issue reported on the [forum](https://forum.guildofwriters.org/viewtopic.php?p=75582#p75582) by denwaler. @Jrius does this look good to you? I know you've mentioned before that the previous Korman behavior was iffy.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -346,7 +346,7 @@ class Exporter:
     def has_coordiface(self, bo):
         if bo.type in {"CAMERA", "EMPTY", "LAMP"}:
             return True
-        if bo.parent is not None:
+        if bo.parent is not None or bo.children:
             return True
         if bo.name in self.actors:
             return True
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -505,9 +505,10 @@ class MeshConverter(_MeshManager):
                     face_verts.append(data.blender2gs[vertex][coluv])
 
             # Convert to triangles, if need be...
-            if len(face_verts) == 3:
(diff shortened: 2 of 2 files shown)


Jrius:
> Apply all transforms if we don't have a CI. Empirical evidence suggests that simply stashing the transform matrices into the spans can be wiped away by plEnableMsg (WTF)

Looks like the draw interface expects matrices from the drawable span to match those from the coordinate interface. Maybe some kind of optimization, or just more weird Plasma behavior.

Anyway. In PotS, I'm noticing issues with unrelated objects now being drawn at an incorrect location (like their transform is applied twice). [Here is an example file](https://www.dropbox.com/s/i08h37tfaigj7lf/monkeyloc.blend?dl=1), the issue is only visible in-game (big parent monkey is drawn at incorrect location, small child monkey is where it should be).

The affected object are those who:
- Are drawable
- AND parent another object (thus have a coordinate interface)
- AND have no collision (apparently, having a collider fixes the issue O.o)

Children objects end up at the correct location, so this really affects only the parent's drawable, not the coordinate interface itself.

I tried comparing this to ALRDoor01 in Ahnonay's Cathedral, which has a similar setup, but couldn't find what's wrong. The coordinate interface/drawable span have no meaningful difference AFAICT.

Hoikas:
Good catch! This exposed a logic issue where non-animated parent objects were not always getting a coordinate interface. That's only been a problem since the beginning, apparently.

Jrius:
Ah, so the transform was still baked in the vertices, that makes sense. That fixes it !

PR #275, opened 2021-08-14
Warn if more than 8 RT lamps are on a material.

TikiBear:
I'm probably late to the party, but I have found that PotS can't handle more than 8 RT lamps on one material. This might vary by target engine, though, so no failure pronouncement is made.

### DIFF
--- a/korman/exporter/rtlight.py
+++ b/korman/exporter/rtlight.py
@@ -258,6 +258,7 @@ class LightConverter:
         # If there is no light group, we'll say that there is no runtime lighting...
         # If there is, we will harvest all Blender lamps in that light group that are Plasma Objects
         lg = bm.light_group
+        lcount = 0
         if lg is not None:
             for obj in lg.objects:
                 if obj.type != "LAMP":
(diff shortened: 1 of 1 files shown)


korman/exporter/rtlight.py
```diff
@@ -291,6 +292,10 @@ def find_material_light_keys(self, bo, bm):
                 else:
                     self._report.msg("[{}] PermaLight '{}'", lamp.type, obj.name, indent=2)
                     permaLights.append(pl_light)
+                    lcount = lcount + 1
+    
+        if lcount > 8:
+            self._report.warn("More than 8 RT lamps on material: '{}'".format(bm.name), indent=1)
```
Hoikas, line 298:
```suggestion

        if len(permaLights) > 8:
            self._report.warn("More than 8 RT lamps on material: '{}'", bm.name, indent=1)
```

korman/exporter/rtlight.py
```diff
@@ -258,6 +258,7 @@ def find_material_light_keys(self, bo, bm):
         # If there is no light group, we'll say that there is no runtime lighting...
         # If there is, we will harvest all Blender lamps in that light group that are Plasma Objects
         lg = bm.light_group
+        lcount = 0
```
Hoikas, line 261:
```suggestion
```

Hoikas (commented):
This seems like a reasonable warning; the 8 RT light limit is an inherent limitation of DirectX's fixed function pipeline, so there is no flexibility on that per engine.

PR #277, opened 2021-08-17
Clearly deliniate the circles of hell.

### LINKED ISSUE #240: Satan unleashed unevenly on shared materials
I have two objects sharing a material, and one is set up with modifiers to unleash Satan (both runtime lighting and preshade material colours). At export, the resulting material will be incorrect for one of those objects depending on the export order.

We probably need to detect and make duplications of materials if they are shared across objects in different circles of hell.

Hoikas:
If we've unleashed Satan on a material, he will perform his dark magick on the material colors. Therefore, we need to prefix the satanic materials with something other than "RTLit_" lest the plain old runtime lit materials be subjugated by the prince of darkness.

Fixes #240

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -198,7 +198,13 @@ class MaterialConverter:
             hgmat = None
         else:
             # Ensure that RT-lit objects don't infect the static-lit objects.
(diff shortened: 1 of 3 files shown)


Deledrius:
🤣

Hoikas:
:trollface:

dpogue (approved):

PR #278, opened 2021-08-18
Don't mark lights in a group as `kLPMovable`.

### LINKED ISSUE #264: Ability to turn off LPMovable flag for animated lamps
By default, all animated lamps are exported with the `kLPMovable` flag, which causes the lamps to be applied to all objects in the scene at runtime.

This poses an issue for objects that are trying to use light groups to restrict light behaviour to specific objects, since the engine ignores the light group at runtime leading to unintended lighting on other objects.

A few potential options:
* Lamps with only colour animations should not be marked as movable
  * Risks causing animated lamps to not properly apply at runtime
  * Does not solve the light group issue for lamps with animated positions
* Lamps that are part of light groups should not be marked as movable
  * Not necessarily intuitive behaviour for artists
* Lamps should have a checkbox option to force not movable
  * Complicates the UI by adding clutter for rarely used options

Hoikas:
It looks like simply attaching a lamp to a group already has fairly major implications for how it's exported. If you're adding a lamp into a group, then you are basically saying, "I only want to affect these specific objects.", so marking them as movable (meaning, "apply me to all objects") doesn't really make sense in that context.

Fixes #264
CC @DoobesURU

### DIFF
--- a/korman/exporter/rtlight.py
+++ b/korman/exporter/rtlight.py
@@ -193,7 +193,7 @@ class LightConverter:
 
         # If the lamp has any sort of animation attached, then it needs to be marked movable.
         # Otherwise, Plasma may not use it for lighting.
-        if bo.plasma_object.has_animation_data:
+        if bo.plasma_object.has_animation_data and not has_lg:
             pl_light.setProperty(plLightInfo.kLPMovable, True)
 
         # *Sigh*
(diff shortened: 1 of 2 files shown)


dpogue (approved):

DoobesURU:
This makes sense to me. IIRC, any lamps that I have that are animated are not part of light groups anyway, so this should be a good sanity check.

Hoikas:
💥

PR #283, opened 2021-08-25
Fix RT lit material shading animations

Hoikas:
Requires corresponding HSPlasma fix (attached for 2.79b).

[2.79.zip](https://github.com/H-uru/korman/files/7050311/2.79.zip)


Hoikas:
:shipit:

PR #290, opened 2021-09-01
Matrix anim fix

### LINKED ISSUE #284: Animated child object improvement observation
While working on Descent, I've noticed that, when there's child objects with animation that are  parented, the "affine parts" of the ATC animation that are exported by Korman are incorrect, specifically the T value, which currently just uses the child object's location in Blender. Here's what I've observed:

* For the top parent object (ie not parented to anything else), the affine parts' **hsVector T value** is the location as it appears in Blender like Korman does now.
* For subsequent child objects that are also animated, the affine parts' **hsVector T value** is the child object's origin's offset distance from the parent's origin.

Sadly, I don't know exactly where to go in the coding to fix all this or to tell Korman to do this only for child objects, so I'm posting the issue here.

Jrius:
Alright, this should mostly fix #284.

What this does:
- Replace all `object.location` with `object.matrix_world_transform`, since `object.location` is not reliable as world transform.
- Replace all `object.matrix_basis` with `object.matrix_local` when we want to access local transform.
- Fix a few smallish issues I noticed while I'm at it (such as `bo.name` in `generate_flat_proxy`).
- Apply `object.matrix_parent_inverse` to animations to _attempt_ to make them truly local.
Emphasis on "attempt", because mathematically, animations in non-local space cannot always be represented via local-space T/R/S anims. Also, wide euler rotations can cause issues due to being converted to quats and back to eulers; and mixing complex scaling/rotation in the parent & child can have some odd results. There is still room for improvement in case more issues crop up: for instance we could use `hsBezScaleKey` or enforce quaternion usage instead of eulers. But currently 80-90% (very rough estimate) of non-local animations should export fine. Local animations will export like they used to, so no problem there.
Also, a message is logged in case the animation is non-local, to help with debugging.

Additionally, this should help fix potential misplacement issues with the following:
- ladders mount direction
- journal/books with no clickregion
- dynamic env map positioning
- water height positioning
- circle cameras
- collisions for swim region attached to basic water

I tested all of them in various configuration and AFAICT all work fine now.

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -342,7 +342,7 @@ class AnimationConverter:
                                      start: Optional[int] = None, end: Optional[int] = None) -> Optional[plMatrixChannelApplicator]:
         if adjust_xform != mathutils.Matrix.Identity(4):
             self._exporter().report.warn(("{}: Transform animation is not local and may export incorrectly. " +
-                "Please use Ctrl-P -> Clear Parent Inverse before animating objects to avoid issues.").format(bo.name), indent=1)
+                "Please use Alt-P -> Clear Parent Inverse before animating objects to avoid issues.").format(bo.name), indent=1)
         else:
             # Adjustment matrix is identity, just pass None instead...
             adjust_xform = None
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -397,7 +397,7 @@ class AnimationConverter:
 
             def convert_scale_keyframe(scale):
                 # Scale: very likely to cause issues.
-                return (a * b for a, b in zip(adjust_scale, scale))
+                return [a * b for a, b in zip(adjust_scale, scale)]
 
             convert_pos = convert_pos_keyframe
             convert_rot = convert_rot_keyframe
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -58,7 +58,7 @@ class AnimationConverter:
         if isinstance(bo.data, bpy.types.Camera):
             applicators.append(self._convert_camera_animation(bo, so, obj_fcurves, data_fcurves, anim_name, start, end))
         else:
-            applicators.append(self._convert_transform_animation(bo, obj_fcurves, bo.matrix_basis, start=start, end=end))
+            applicators.append(self._convert_transform_animation(bo, obj_fcurves, bo.matrix_local, bo.matrix_parent_inverse, start=start, end=end))
         if bo.plasma_modifiers.soundemit.enabled:
             applicators.extend(self._convert_sound_volume_animation(bo.name, obj_fcurves, bo.plasma_modifiers.soundemit, start, end))
         if isinstance(bo.data, bpy.types.Lamp):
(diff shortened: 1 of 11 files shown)


korman/exporter/physics.py
```diff
@@ -93,7 +93,7 @@ def generate_flat_proxy(self, bo, so, **kwargs):
 
         if so.sim is None:
             simIface = self._mgr.add_object(pl=plSimulationInterface, bl=bo)
-            physical = self._mgr.add_object(pl=plGenericPhysical, bl=bo, name=name)
+            physical = self._mgr.add_object(pl=plGenericPhysical, bl=bo, name=bo.name)
```
Hoikas, line 96: The name argument could probably just be removed. :trollface:

korman/exporter/animation.py
```diff
@@ -351,20 +358,62 @@ def _convert_transform_animation(self, bo, fcurves, xform, *, allow_empty: Optio
...
+                else: # tuple
+                    if len(rot) == 4: # quat in a tuple
+                        return (adjust_rotation * mathutils.Quaternion(rot))[:]
+                    else: # XYZ euler in a tuple
+                        rot = mathutils.Euler(rot, "XYZ").to_quaternion()
+                        return (adjust_rotation * rot).to_euler("XYZ")[:]
+
+            def convert_scale_keyframe(scale):
+                # Scale: very likely to cause issues.
+                return (a * b for a, b in zip(adjust_scale, scale))
```
Hoikas, line 400:
```suggestion
                return [a * b for a, b in zip(adjust_scale, scale)]
```

I think it would be best to use a list comprehension here. The underlying keyframe processing code expects sequences, not generators 😉

DoobesURU:
Fantastic! As soon as I get a chance, I'll test these changes on some of my projects.

Hoikas (commented):
Whew, thanks for looking into this. I think I would like to defer this until after Korman 0.12 releases - which is "just as soon as I can push the tag", then we can poke at this a bit more 😄

Jrius:
> I think I would like to defer this until after Korman 0.12 releases - which is "just as soon as I can push the tag", then we can poke at this a bit more

Understood ! ;)

Hoikas (approved):

Hoikas:
@DoobesURU Any news on running this through your test suite?

korman/exporter/animation.py
```diff
@@ -338,9 +338,16 @@ def convert_spot_inner(spot_blend, spot_size):
             applicator.channel = channel
             yield applicator
 
-    def _convert_transform_animation(self, bo, fcurves, xform, *, allow_empty: Optional[bool] = False,
+    def _convert_transform_animation(self, bo, fcurves, default_xform, adjust_xform, *, allow_empty: Optional[bool] = False,
                                      start: Optional[int] = None, end: Optional[int] = None) -> Optional[plMatrixChannelApplicator]:
-        tm = self.convert_transform_controller(fcurves, bo.rotation_mode, xform, allow_empty=allow_empty,
+        if adjust_xform != mathutils.Matrix.Identity(4):
+            self._exporter().report.warn(("{}: Transform animation is not local and may export incorrectly. " +
+                "Please use Ctrl-P -> Clear Parent Inverse before animating objects to avoid issues.").format(bo.name), indent=1)
```
DoobesURU, line 345: This is actually "Alt+P". 😉

DoobesURU:
> 
> 
> @DoobesURU Any news on running this through your test suite?

I've run this through the only Age I have that has parented animated objects: Descent. After fixing things so all the working

> 
> 
> @DoobesURU Any news on running this through your test suite?

So after running the only Age I have that has parented animated objects through the gauntlet (Descent) and after some adjustments, it looks like things are working as they should as far as child objects aligning correctly and not having things off in space. FWIW, I give the PR my seal of approval. 😃

PR #293, opened 2021-09-09
Code that kills opacity animations

TikiBear:
Just noting code that kills opacity animations, triggering message, "No controllable animations were found." I'm guessing that just removing the code isn't the answer, though doing so restores opacity animations without obvious harm in my limited testing (Flying Cauldron). I don't know how to figure this one out, unfortunately.

Per ametist, note that this impacts Tiam.

### DIFF
--- a/korman/nodes/node_messages.py
+++ b/korman/nodes/node_messages.py
@@ -296,7 +296,8 @@ class PlasmaAnimCmdMsgNode(idprops.IDPropMixin, PlasmaMessageWithCallbacksNode, 
                 self.raise_error("At least one of: target object, material, texture MUST be specified")
             target = exporter.mesh.material.get_texture_animation_key(obj, material, texture, self.anim_name)
 
-        target = [i for i in target if not isinstance(i.object, (plAgeGlobalAnim, plLayerSDLAnimation))]
+        # this kills opacity animation:
+        # target = [i for i in target if not isinstance(i.object, (plAgeGlobalAnim, plLayerSDLAnimation))]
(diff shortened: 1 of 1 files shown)


korman/nodes/node_messages.py
```diff
@@ -296,7 +296,8 @@ def convert_message(self, exporter, so):
                 self.raise_error("At least one of: target object, material, texture MUST be specified")
             target = exporter.mesh.material.get_texture_animation_key(obj, material, texture, self.anim_name)
 
-        target = [i for i in target if not isinstance(i.object, (plAgeGlobalAnim, plLayerSDLAnimation))]
+        # this kills opacity animation:
+        # target = [i for i in target if not isinstance(i.object, (plAgeGlobalAnim, plLayerSDLAnimation))]
```
Hoikas, line 300:
```suggestion
        target = [i for i in target if isinstance(i.object, (plAgeGlobalAnim, plLayerSDLAnimation))]
```

Hoikas (changes requested):
It looks like the real problem is that the condition may be inverted.

Hoikas:
Actually, no, the condition is correct.

Hoikas:
Korman 0.12 tightens up the detection of animations that are addressable via nodes. Previously, Korman would just export what it thought was a reference to the correct animation by name. But, this was error prone, and it might generate references to animations that do not exist. This has been tightened up in Korman 0.12, and, as the error message states, the combination that you've picked in the node doesn't match an animation.

Issue #136, opened 2019-02-01
GUI Dialog Nodes


Hoikas:
This should be the start of the revamp of the concept of a "Page" in Korman. Most pages will be what they currently are: age geometry. However, some pages should never contain age geometry. Specifically, avatar animation pages and GUI pages. Recall that each avatar animation and GUI dialog MUST be its own page.

Therefore, the basic idea is to have a page type that allows a node tree to be specified. That tree is where we spec the GUI Dialog. A modifier should be added for simpler GUI elements, like an image popup.

This is blocking on #140.

Hoikas:
Deferring to later release, potentially 0.11.

Hoikas:
Deferring to v0.14 due to scope change of v0.13.

Hoikas:
Clearing out milestone for now because it just keeps getting deferred.

Hoikas:
As of #373, I am going to call this issue complete. While we do not support all GUI controls by far, the basics are in. Future work should be guided by more specific requests (eg requests for specific controls).

PR #294, opened 2021-09-09
Add Custom GUI Dialog Modifier

DoobesURU:
Adds the beginnings of the basic custom GUI dialog modifier.

The goal will be to eventually create a simple GUI Dialog mod for single images per #136 , along with the required Post Effects Mod plus (optional) GUI Button for clicking off.

Along with the basic components, there is a crude sanity check to make sure there's at least a clickable and the GUI object itself, plus a pre-export def to check for the version(s) for export.

The modifier itself indicates what is required and what is optional, but of course, the final layout is up for suggestions.

It still needs the actual export component and probably a few other things. Presumably, there's things that need to be done so we don't need to mirror the GUI object's texture backwards or rotate the camera (empty?) to face away after setup.

At the end, we need a GUI Dialog Mod, a Post Effects Mod, an (optional) GUI Button Mod, and a Python script reference for xDialogToggle.py, everything with all the trimmings.

I'm also not sure if we want this to be more broad and use the opportunity to also work on the aforementioned dialog nodes/PRPs detailed in the referenced issue.

Also, minor issue: The name/label is cut off as "Dialog" in the modifier drop-down menu.

### DIFF
--- a/korman/exporter/manager.py
+++ b/korman/exporter/manager.py
@@ -27,6 +27,8 @@ from ..plasma_magic import *
 # NOTE: We are using Factory indices because I doubt all of these classes are implemented.
 _pool_types = (
     plFactory.ClassIndex("plPostEffectMod"),
+    plFactory.ClassIndex("plSingleModifier"),
+    plFactory.ClassIndex("pfGUIControlMod"),
     plFactory.ClassIndex("pfGUIDialogMod"),
     plFactory.ClassIndex("pfGUIButtonMod"),
     plFactory.ClassIndex("plMsgForwarder"),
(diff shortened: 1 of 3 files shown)


DoobesURU:
I've updated a few things with new knowledge from my dealings with the Linking Book mod. I'm not entirely sure what goes into the `def export`, but I'm assuming it'll need `plDialogMod` and `plButtonMod` at the very least.

Also, I believe there was talk about pushing the GUI onto a new page? If so, what would be the best way to approach that?

DoobesURU:
Making some progress with export (I think). Current hitch is getting `kModal` to toggle. It can't seem to find the attribute in `pfGUIDialogMod` no matter what combination I put in.

Hoikas:
Closing in favor of #373.

PR #295, opened 2021-09-13
Add Local Only to Sound Emitter Modifier

DoobesURU:
Adds the "Local Only/kIsLocalOnly" property to sound emitter modifiers. Dependent on the kPropIsLocalOnly attribute, which still needs to be added to libHSPlasma (unless I'm using the wrong definition?).

This will allow better sound control per client/avatar, which Ages like Tiam and Descent will need.

### DIFF
--- a/korman/properties/modifiers/sound.py
+++ b/korman/properties/modifiers/sound.py
@@ -373,7 +373,12 @@ class PlasmaSound(idprops.IDPropMixin, bpy.types.PropertyGroup):
         if self.incidental:
             sound.properties |= plSound.kPropIncidental
         if self.local_only:
-            sound.properties |= plSound.kPropLocalOnly
+            # Local only and random sounds hate each other, so we must check for both
+            for i in bpy.data.objects:
+                if i.plasma_modifiers.random_sound:
+                    raise ExportError("SoundEmitter '{}': cannot have local only sounds and a random sound mod.".format(self.id_data.name))
+                else:
+                    sound.properties |= plSound.kPropLocalOnly
         sound.dataBuffer = self._find_sound_buffer(exporter, so, wavHeader, dataSize, channel)
 
         # Cone effect
--- a/korman/properties/modifiers/sound.py
+++ b/korman/properties/modifiers/sound.py
@@ -243,6 +243,10 @@ class PlasmaSound(idprops.IDPropMixin, bpy.types.PropertyGroup):
                         description="Loop the sound",
                         default=False,
                         options=set())
+    local_only = BoolProperty(name"Local Only",
+                              description="Sounds only plays for local avatar",
+                              default=False,
+                              options=set())
 
     inner_cone = FloatProperty(name="Inner Angle",
                                description="Angle of the inner cone from the negative Z-axis",
@@ -368,6 +372,8 @@ class PlasmaSound(idprops.IDPropMixin, bpy.types.PropertyGroup):
             sound.properties |= plSound.kPropLooping
         if self.incidental:
             sound.properties |= plSound.kPropIncidental
+        if self.local_only:
+            sound.properties |= plSound.kPropIsLocalOnly
         sound.dataBuffer = self._find_sound_buffer(exporter, so, wavHeader, dataSize, channel)
 
         # Cone effect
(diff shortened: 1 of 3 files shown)


korman/properties/modifiers/sound.py
```diff
@@ -368,6 +372,8 @@ def _convert_sound(self, exporter, so, pClass, wavHeader, dataSize, channel=None
             sound.properties |= plSound.kPropLooping
         if self.incidental:
             sound.properties |= plSound.kPropIncidental
+        if self.local_only:
+            sound.properties |= plSound.kPropIsLocalOnly
```
dpogue, line 376: There is an existing `kPropLocalOnly` that should do what you want here: https://github.com/H-uru/libhsplasma/blob/35b100f64d6fe54479737e885d9809fcebafdf86/core/PRP/Audio/plSound.h#L35

DoobesURU: Ah, so I got the property name wrong. I'll try that and let you know.

DoobesURU:
@Hoikas , I need your opinion about modifying the sound message node to add the local only option:

Do we want the node to simply inherit the local only status from the modifier and pass it along for export or do we want to give the option to change it like the other settings? If it's the latter, do we want an EnumProperty like the rest or a simple BoolProperty?

Hoikas:
It looks like you can't change the entire local only property of a sound from a message. You can, however, set a sound message to only perform its action in response to local notifications - which is slightly different. I'd say hold off on doing anything in node land for now.

DoobesURU:
> 
> 
> It looks like you can't change the entire local only property of a sound from a message. You can, however, set a sound message to only perform its action in response to local notifications - which is slightly different. I'd say hold off on doing anything in node land for now.

I see where you're coming from. I ask because I _think_ the kPropLocalOnly in the sound emitter properties and the kIsLocalOnly command in the sound message work in tandem, or at least that's how it appears to be in Cyan's files.

So I'm guessing the former option...the message node inherits what the modifier says...would be the wisest course of action on this? As it stands now, the kPropLocalOnly part works at least (thanks to @dpogue pointing me the right way), so we can leave it at that if you want to wait on modifying the node.

DoobesURU:
I noticed in testing that local only does not go along with random sounds, so I put in an error message if that happens. @Hoikas , anything else I should be poking at for this?

korman/properties/modifiers/sound.py
```diff
@@ -368,6 +372,13 @@ def _convert_sound(self, exporter, so, pClass, wavHeader, dataSize, channel=None
             sound.properties |= plSound.kPropLooping
         if self.incidental:
             sound.properties |= plSound.kPropIncidental
+        if self.local_only:
+            # Local only and random sounds hate each other, so we must check for both
+            for i in bpy.data.objects:
+                if i.plasma_modifiers.random_sound:
+                    raise ExportError("SoundEmitter '{}': cannot have local only sounds and a random sound mod.".format(self.id_data.name))
+                else:
+                    sound.properties |= plSound.kPropLocalOnly
```
Hoikas, line 381: This will raise an error if any object in the entire blend has a random sound modifier, which is not what you're intending, I suspect. What problem are you noticing with local only on random sounds?

DoobesURU: I think it's only if the random sound mod is referenced in a responder. IIRC, it causes URU to crash...and the dump mentioned that plRandomSoundMod doesn't have a local only reference (or it's named something different). I'll have to roll back and double check that though as the details are hazy at the moment.

If you look at line 375 above that selection, it only checks for the random sound mod if the sound(s) have local only enabled. I've tried exporting Direbo with the check and it exports the random sound mod properly if there's no local only sounds and raises the error if there is one or more within the set.

DoobesURU: After reverting and trying again, I can't seem to reproduce the error (if there even was one related to this PR; might have gotten things mixed up), so I've reverted the local only/random sound check.

Hoikas (commented):
Can we hold off on the changes to the sound message node? I feel like this isn't the correct approach for them - IIRC the `kIsLocalOnly` part of the message indicates a particular command is local only, not that the sound is local only.

DoobesURU:
> Can we hold off on the changes to the sound message node? I feel like this isn't the correct approach for them - IIRC the `kIsLocalOnly` part of the message indicates a particular command is local only, not that the sound is local only.

Done. `node_messages.py` reverted.

Hoikas (approved):

PR #297, opened 2021-10-01
Fix typos in README.

Deledrius:


README.md
```diff
@@ -6,7 +6,7 @@ and its open source variant, CyanWorlds.com Engine. For more information, visit
 
 Dependencies
 ------------
-- [Blender](http://blender3d.org) - 3d modelling problem
+- [Blender](http://blender3d.org) - 3D modeling software
```
dpogue, line 9: In _proper_ English, "modelling" is spelt with 2 Ls :trollface:

Deledrius: It's true.  It's one of those words that frustrates me with being partially incorrect in both forms, according to different rules... ;)

Do we have a standard preference for which English to use?  Despite my inclination, I assume en-US is consistent with most of the codebase.

Hoikas: We prefer Southeastern US English in Korman. :wink:

PR #299, opened 2021-10-02
Add named animation EnumProperty

DoobesURU:
Adds a needed animation picker for new named animations for the Python script texture node. Code borrowed from node_messages.py


DoobesURU:
I've gotten things to the point where the EnumProperty shows up on the node, but the animation(s) do not. I also tried changing anim_name to a PointerProperty for consistency, but that seemed to blow up the node and Korman.

PR #304, opened 2021-11-18
Allow per-texture emit.

### LINKED ISSUE #303: Emissive by texture instead of material.
Currently, Korman exports all textures in a material as emissive when it has `bm.emit > 0`.

What's needed for some Ages, such as K'veer and its ornate pillars, is to make the emit by texture instead.

I propose switching from using `if bm.emit > 0.0:` to using `if slot.use_map_emit:` to toggle emissive by texture. I've tried working the fix into the scripts myself, but it gets rather involved when adding `slot` parameters to the various `defs` and I'm not sure if I'm ripping things apart too much and if there's an easier solution. We can still use the int value from `bm.emit` for the amount of emissive light, which is how ZLZ imports it anyway.

Hoikas:
Fixes #303.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -271,7 +271,7 @@ class MaterialConverter:
         if not hsgmat.layers:
             layer = self._mgr.find_create_object(plLayer, name="{}_AutoLayer".format(mat_name), bl=bo)
             self._obj2layer[bo][bm][None].append(layer.key)
-            self._propagate_material_settings(bo, bm, layer)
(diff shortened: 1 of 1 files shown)


DoobesURU:
Ran this through the K'veer gauntlet and after some back and forth with @Hoikas , this looks to be working so far. I'm guessing a function to use `slot.emit_factor` is in the works as well.

Hoikas:
Merging since it apparently works...

PR #308, opened 2021-12-21
Sounds: allow paths relative to the Blender file.

Jrius:
Makes Korman aware of relative filepaths such as `//sounds/my_sound.ogg` for sounds.

### DIFF
--- a/korman/properties/modifiers/sound.py
+++ b/korman/properties/modifiers/sound.py
@@ -423,7 +423,7 @@ class PlasmaSound(idprops.IDPropMixin, bpy.types.PropertyGroup):
         if sound.packed_file is None:
             stream = hsFileStream()
             try:
-                stream.open(sound.filepath, fmRead)
+                stream.open(bpy.path.abspath(sound.filepath), fmRead)
             except IOError:
                 self._raise_error("failed to open file")
         else:


korman/properties/modifiers/sound.py
```diff
@@ -423,7 +423,7 @@ def _get_sound_info(self):
         if sound.packed_file is None:
             stream = hsFileStream()
             try:
-                stream.open(sound.filepath, fmRead)
+                stream.open(bpy.path.abspath(sound.filepath), fmRead)
```
Hoikas, line 426: Will this still work if the path is already some valid absolute path?

Jrius: Yes. According to the [documentation](https://docs.blender.org/api/blender_python_api_current/bpy.path.html#bpy.path.abspath)  and a few tests, `abspath` does nothing aside from replacing the "//" pattern.

```
>>> bpy.path.abspath("C:/Users/MyUser/Desktop/myfile.txt")
'C:/Users/MyUser/Desktop/myfile.txt'

>>> bpy.path.abspath("C:\\Users\\MyUser\\Desktop\\myfile.txt")
'C:\\Users\\MyUser\\Desktop\\myfile.txt'

>>> bpy.path.abspath("folder/myfile.txt")
'folder/myfile.txt'

>>> bpy.path.abspath("./folder/myfile.txt")
'./folder/myfile.txt'

>>> bpy.path.abspath("//folder/myfile.txt") # assuming current Blender file is saved on the desktop
'C:\\Users\\MyUser\\Desktop\\folder/myfile.txt'
```
Blender itself doesn't usually mix forward and backward slashes, but even if that were the case, `hsFileStream.open` appears to be okay with it on Windows. Should be fine.

Hoikas (approved):

PR #315, opened 2022-02-10
Fix Waveset Amplitude

DoobesURU:
Changes amplitude property from int percentage to a three digit float for more precision, which fixes "wobbly" geostate waves.

### DIFF
--- a/korman/properties/modifiers/water.py
+++ b/korman/properties/modifiers/water.py
@@ -424,7 +424,7 @@ class PlasmaWaveState:
         cls.amplitude = FloatProperty(name="Amplitude",
                                     description="Multiplier for wave height",
                                     precision=3,
-                                    min=0.001, max=1.000, # Minimum must be above zero for shores to work!
+                                    min=0.001, max=1.000, # Minimum must be above zero to prevent rendering problems!
                                     default=cls._amplitude_default)
         cls.chop = IntProperty(name="Choppiness",
                                description="Sharpness of wave crests",
--- a/korman/properties/modifiers/water.py
+++ b/korman/properties/modifiers/water.py
@@ -399,7 +399,7 @@ class PlasmaWaveState:
     def convert_wavestate(self, state):
         state.minLength = self.min_length
         state.maxLength = self.max_length
-        state.ampOverLen = self.amplitude / 100.0
+        state.ampOverLen = self.amplitude
         state.chop = self.chop / 100.0
         state.angleDev = self.angle_dev
 
@@ -407,7 +407,7 @@ class PlasmaWaveState:
         cls = self.__class__
         state.minLength = cls._min_length_default
         state.maxLength = cls._max_length_default
-        state.ampOverLen = cls._amplitude_default / 100.0
+        state.ampOverLen = cls._amplitude_default
         state.chop = cls._chop_default / 100.0
         state.angleDev = cls._angle_dev_default
 
@@ -421,10 +421,10 @@ class PlasmaWaveState:
                                        description="Largest wave length",
                                        min=0.1, max=50.0,
                                        default=cls._max_length_default)
-        cls.amplitude = IntProperty(name="Amplitude",
+        cls.amplitude = FloatProperty(name="Amplitude",
                                     description="Multiplier for wave height",
-                                    subtype="PERCENTAGE",
-                                    min=0, max=100,
+                                    precision=3,
+                                    min=0.001, max=1.000, # Minimum must be above zero for shores to work!
                                     default=cls._amplitude_default)
         cls.chop = IntProperty(name="Choppiness",
                                description="Sharpness of wave crests",
@@ -446,7 +446,7 @@ class PlasmaWaveGeoState(PlasmaWaveState, PlasmaModifierProperties):
 
     _min_length_default = 4.0
     _max_length_default = 8.0
-    _amplitude_default = 10
+    _amplitude_default = 0.001
     _chop_default = 50
     _angle_dev_default = math.radians(20.0)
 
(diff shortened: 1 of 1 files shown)


DoobesURU:
Thanks to ametist for pointing out that a geostate amplitude of zero causes shores to not appear. That's why the minimum is set at 0.001 (per the comment in the script).

korman/properties/modifiers/water.py
```diff
@@ -421,10 +421,10 @@ def register(cls):
                                        description="Largest wave length",
                                        min=0.1, max=50.0,
                                        default=cls._max_length_default)
-        cls.amplitude = IntProperty(name="Amplitude",
+        cls.amplitude = FloatProperty(name="Amplitude",
                                     description="Multiplier for wave height",
-                                    subtype="PERCENTAGE",
-                                    min=0, max=100,
+                                    precision=3,
+                                    min=0.001, max=1.000, # Minimum must be above zero for shores to work!
```
Hoikas, line 427: Will this require Age creators to fix their amplitude values when they upgrade Korman?

Hoikas (commented):
I wonder if some extra safety should be added in such that the exporter code clamps the minimum value of the amplitude to 0.001 if the Water Shore modifier is enabled?

korman/properties/modifiers/water.py
DoobesURU: Things seemed to carry over fine on both my Noloben and Laki blends, so no, I don't believe so

DoobesURU: And yeah, I was thinking about the clamp for when the shores mod is active, but was not entirely sure how to do that. The UI scripts are hit or miss when I modify them. 😊

dpogue:
Wavesets with 0 amplitude were also causing issues in shaders for macOS and OpenGL. There's a point where it divides by the amplitude, and gets unhappy dividing by 0. DirectX seems to clamp the result to 0 instead of infinity.

> **[maccodemonkey]**	I was having trouble with wave sets tonight and the newer version of Veelay Tsahvahn I have. Was just wondering about reconilation in since I'm not sure if the issue is fixed in the latest. Some of the wave set data is getting initialized to inf inside of the waveset class, and I haven't been touching that.
> **[maccodemonkey]**	On the 2020 era game data I have on my other machine, the waters rendering fine, so my guess has been that something changed in the engine that the updated age is expecting.
> **[dpogue]**	hmmm, I remember https://github.com/H-uru/Plasma/pull/977 but I think that was mostly a theoretical issue
> **[maccodemonkey]**	yeah, I didn't see anything that looked relevant yet in history. Still debugging. Could be a difference in how the AMD GPU on this Mac handles inf values inside the shader. fAmpOverLen is getting set to inf, which then gets into normQ, which then leaks into the shader, which then causes all the vertices to be NaN
> **[dpogue]**	AmpOverLen should be coming from the data, shouldn't it?
> **[maccodemonkey]**	thats what I was just looking at.
> **[maccodemonkey]**	oh I see, normQ is getting filled in as 0/0 which is inf, ampOverLen is 0, not inf. I misspoke, I thought I saw it as inf.
> **[maccodemonkey]**	oh, 0.25/0, which is still inf
> **[dpogue]**	ampoverlen being 0 seems like bad settings in the data, but should maybe be handled as normQ being 0?

DoobesURU:
> Wavesets with 0 amplitude were also causing issues in shaders for macOS and OpenGL. There's a point where it divides by the amplitude, and gets unhappy dividing by 0. DirectX seems to clamp the result to 0 instead of infinity.

Ah, so another reason to keep amplitude > 0 at all times.

korman/properties/modifiers/water.py
```diff
@@ -446,7 +446,7 @@ class PlasmaWaveGeoState(PlasmaWaveState, PlasmaModifierProperties):
 
     _min_length_default = 4.0
     _max_length_default = 8.0
-    _amplitude_default = 10
+    _amplitude_default = 0.001
```
Hoikas, line 449: Shouldn't this be `0.1`?

korman/properties/modifiers/water.py
```diff
@@ -464,7 +464,7 @@ class PlasmaWaveTexState(PlasmaWaveState, PlasmaModifierProperties):
 
     _min_length_default = 0.1
     _max_length_default = 4.0
-    _amplitude_default = 10
+    _amplitude_default = 0.040
```
Hoikas, line 467: `0.1`?

korman/properties/modifiers/water.py
DoobesURU: Most wavesets I've seen, in Cyan's work, in the old AlcScript, etc, the geostate's amplitude was quite small, usually around 0.0019953...etc. The reason some have been getting "wobbly" waveset geometry (including me) was because, even at 1%/0.01, that was set too high.

korman/properties/modifiers/water.py
DoobesURU: This setting was around average of what I've seen in most of the wavesets. This default creates a rather nice ripple effect (such as the ocean in Noloben) that Age builders can then play around with from there.

DoobesURU:
These changes are based on some of my own observations when playing with waveset settings both in Korman and in PRPShop. Changing this to a float will also be helpful to those coming from PyPRP as this is what the original AlcScript displayed it as as well: https://www.guildofwriters.org/wiki/Wavesets#The_Parameters

korman/properties/modifiers/water.py
DoobesURU: Looking at the old AlcScript for wavesets (yeah, I went a bit cross-eyed looking at *that* stuff again 😜 ), an argument could be made for setting the default at 0.010.

korman/properties/modifiers/water.py
Hoikas: Ok, I just wasn't expecting to see the default values to change (except to match the calculation changes made in the export routine) based off the PR description.

Hoikas (approved):

PR #317, opened 2022-02-24
Add Share Function to Linking Book Mod

DoobesURU:
Adds a "Shareable" toggle to Linking Book Mod when Link Type is set to Private Link (MOUL only).

### DIFF
--- a/korman/properties/modifiers/gui.py
+++ b/korman/properties/modifiers/gui.py
@@ -601,7 +601,7 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         responder.link_output(responder_state, "state_refs", "resp")
         responder.link_output(linkingnode, "keyref", "respOneShot")
 
-    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region, share_rgn):
+    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region, share_region):
         # Clickable
         clickable_region = nodes.new("PlasmaClickableRegionNode")
         clickable_region.region_object = clk_region
@@ -659,11 +659,11 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         # Share MSB
         if share_region is not None:
             # Region
-            share_region = nodes.new("PlasmaVolumeSensorNode")
-            share_region.region_object = share_rgn
-            for i in share_region.inputs:
+            share_msb_region = nodes.new("PlasmaVolumeSensorNode")
+            share_msb_region.region_object = share_region
+            for i in share_msb_region.inputs:
                 i.allow = True
-            share_region.link_output(linkingnode, "satisfies", "shareRegion")
+            share_msb_region.link_output(linkingnode, "satisfies", "shareRegion")
             # MSB Behavior
             share_seek = nodes.new("PlasmaSeekTargetNode")
             share_seek.target = self.seek_point
--- a/korman/properties/modifiers/gui.py
+++ b/korman/properties/modifiers/gui.py
@@ -482,19 +482,22 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         else:
             rgn_obj = self.clickable_region
 
-        # Generate the same six foot cube if we need a share region.
         if self.shareable:
             if self.share_region is None:
                 with utils.bmesh_object("{}_LinkingBook_ShareRgn".format(self.key_name)) as (share_region, bm):
+                    # Generate a cube for the share region.
                     bmesh.ops.create_cube(bm, size=(8.0))
                     share_region_offset = mathutils.Matrix.Translation(self.clickable.matrix_world.translation - share_region.matrix_world.translation)
                     bmesh.ops.transform(bm, matrix=share_region_offset, space=share_region.matrix_world, verts=bm.verts)
                     share_region.plasma_object.enabled = True
                     share_region.hide_render = True
+                    yield share_region
             else:
                 share_region = self.share_region
+        else:
+            share_region = None
 
-            yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), click_region=rgn_obj, share_region=share_region)
+        yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), click_region=rgn_obj, share_region=share_region)
 
     def export(self, exporter, bo, so):
         if self._check_version(pvPrime, pvPots):
@@ -598,7 +601,7 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         responder.link_output(responder_state, "state_refs", "resp")
         responder.link_output(linkingnode, "keyref", "respOneShot")
 
-    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region, shr_region):
+    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region, share_rgn):
         # Clickable
         clickable_region = nodes.new("PlasmaClickableRegionNode")
         clickable_region.region_object = clk_region
@@ -654,10 +657,10 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
         linking_panel_name.link_output(linkingnode, "pfm", "TargetAge")
 
         # Share MSB
-        if self.shareable:
+        if share_region is not None:
             # Region
             share_region = nodes.new("PlasmaVolumeSensorNode")
-            share_region.region_object = shr_region
+            share_region.region_object = share_rgn
             for i in share_region.inputs:
                 i.allow = True
             share_region.link_output(linkingnode, "satisfies", "shareRegion")
--- a/korman/properties/modifiers/gui.py
+++ b/korman/properties/modifiers/gui.py
@@ -427,6 +427,10 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
                                 description="The object the avatar will click on to activate the Linking Book GUI",
                                 type=bpy.types.Object,
                                 poll=idprops.poll_mesh_objects)
+    shareable = BoolProperty(name="Shareable",
+                             description="Enable the Book to be Shareable (MOUL private instance only)",
+                             default=False,
+                             options=set())
 
     # -- Path of the Shell options --
     # Popup Appearance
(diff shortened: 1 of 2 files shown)


DoobesURU:
At the moment, for sanity's sake, the Shareable toggle only appears when MOUL is selected and the Link Type is private instanced. Thinking on this more, does an extra sanity check to filter out ABM and PotS-only Ages need to be put in or will Korman do that anyway?

korman/properties/modifiers/gui.py
```diff
@@ -427,6 +427,14 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
                                 description="The object the avatar will click on to activate the Linking Book GUI",
                                 type=bpy.types.Object,
                                 poll=idprops.poll_mesh_objects)
+    shareable = BoolProperty(name="Shareable",
+                             description="Enable the Book to be Shareable (MOUL private instance only)",
+                             default=False,
+                             options=set())
+    shr_region = PointerProperty(name="Share Region (optional)",
```
Hoikas, line 434: I think this should be `share_region`. I think I would really have trouble remembering this abbreviation ;)

korman/properties/modifiers/gui.py
```diff
@@ -633,6 +641,34 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
         linking_panel_name.value = self.link_destination if self.link_destination else self.age_name
         linking_panel_name.link_output(linkingnode, "pfm", "TargetAge")
 
+        # Share MSB
+        if self.shareable:
+        # Region
```
Hoikas, line 646:
```suggestion
            # Region
```

korman/properties/modifiers/gui.py
```diff
@@ -633,6 +641,34 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
...
+            share_region = nodes.new("PlasmaVolumeSensorNode")
+            if self.shr_region is None:
+                share_region.region_object = clk_region
+            else:
+                share_region.region_object = self.shr_region
+            share_region.bounds = "hull"
```
Hoikas, line 652: Don't hardcode this - the physics exporter will grab the settings from the collision modifier, if it's enabled. We don't want to clobber any user settings :)

korman/properties/modifiers/gui.py
```diff
@@ -633,6 +641,34 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
...
+            for i in share_region.inputs:
+                i.allow = True
+            share_region.link_output(linkingnode, "satisfies", "shareRegion")
+        # MSB Behavior
```
Hoikas, line 656:
```suggestion
            # MSB Behavior
```

korman/properties/modifiers/gui.py
```diff
@@ -427,6 +427,14 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
                                 description="The object the avatar will click on to activate the Linking Book GUI",
                                 type=bpy.types.Object,
                                 poll=idprops.poll_mesh_objects)
+    shareable = BoolProperty(name="Shareable",
+                             description="Enable the Book to be Shareable (MOUL private instance only)",
+                             default=False,
```
Hoikas, line 432: I think this should probably default to `True` - I can't think of a reason why we want private books to be non-sharable by default.

Hoikas (commented):
This also needs to explicitly export the share region under the `pre_export` function. You might want to copy what is being done for `clickable_region` so that a share region will automatically be generated.

korman/properties/modifiers/gui.py
DoobesURU: Agreed. Only issue I see is it'll overlap with the `share_region` definition of the new node (line 647). I can rename that share_region_msb to differentiate.

Hoikas: Having the same name would be a good thing, IMO 😄

DoobesURU: Heh, OK. I thought it'd cause problems, but I defer to your expertise as always. 😃

DoobesURU:
Implemented all the fixes you mentioned, @Hoikas !

korman/ui/modifiers/gui.py
```diff
@@ -88,6 +88,9 @@ def row_alert(prop_name, **kwargs):
 
         layout.separator()
         layout.prop(modifier, "link_type")
+        if modifier.link_type == "kOriginalBook":
+            layout.prop(modifier, "shareable")
+            layout.prop(modifier, "shr_region")
```
Hoikas, line 93:
```suggestion
            layout.prop(modifier, "share_region")
```

korman/properties/modifiers/gui.py
```diff
@@ -427,6 +427,14 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
...
                                 type=bpy.types.Object,
                                 poll=idprops.poll_mesh_objects)
+    shareable = BoolProperty(name="Shareable",
+                             description="Enable the Book to be Shareable (MOUL private instance only)",
+                             default=True,
+                             options=set())
+    share_region = PointerProperty(name="Share Region (optional)",
+                                       description="Sets an alternate share region (optional)",
+                                       type=bpy.types.Object,
+                                       poll=idprops.poll_mesh_objects)
```
Hoikas, line 437: Looks like the indentation needs to be fixed here.

Hoikas (commented):
Don't forget about the `.pre_export()` thing for the share region ;)

korman/ui/modifiers/gui.py
DoobesURU: Something I _always_ tend to forget. 😜 Fixed.

korman/properties/modifiers/gui.py
DoobesURU: That ugly whitespace! It's gone now.

korman/properties/modifiers/gui.py
```diff
@@ -474,7 +482,19 @@ def pre_export(self, exporter, bo):
...
+                with utils.bmesh_object("{}_LinkingBook_ShareRgn".format(self.key_name)) as (share_region, bm):
+                    bmesh.ops.create_cube(bm, size=(8.0))
+                    share_region_offset = mathutils.Matrix.Translation(self.clickable.matrix_world.translation - share_region.matrix_world.translation)
+                    bmesh.ops.transform(bm, matrix=share_region_offset, space=share_region.matrix_world, verts=bm.verts)
+                    share_region.plasma_object.enabled = True
+                    share_region.hide_render = True
+            else:
+                share_region = self.share_region
+
+            yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), click_region=rgn_obj, share_region=share_region)
```
Hoikas, line 497:
```suggestion
        yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), click_region=rgn_obj, share_region=share_region)
```

We need to be sure that we convert the logic even if the book is not sharable ;)

korman/properties/modifiers/gui.py
```diff
@@ -633,6 +653,30 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
         linking_panel_name.value = self.link_destination if self.link_destination else self.age_name
         linking_panel_name.link_output(linkingnode, "pfm", "TargetAge")
 
+        # Share MSB
+        if self.shareable:
```
Hoikas, line 657:
```suggestion
        if share_region is not None:
```

korman/properties/modifiers/gui.py
```diff
@@ -633,6 +653,30 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
...
+            # Region
+            share_region = nodes.new("PlasmaVolumeSensorNode")
```
Hoikas, line 659: This will collide with the function argument named `share_region`. Whoops!

korman/properties/modifiers/gui.py
```diff
@@ -474,7 +482,19 @@ def pre_export(self, exporter, bo):
...
-        yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), region=rgn_obj)
+        # Generate the same six foot cube if we need a share region.
+        if self.shareable:
+            if self.share_region is None:
+                with utils.bmesh_object("{}_LinkingBook_ShareRgn".format(self.key_name)) as (share_region, bm):
+                    bmesh.ops.create_cube(bm, size=(8.0))
+                    share_region_offset = mathutils.Matrix.Translation(self.clickable.matrix_world.translation - share_region.matrix_world.translation)
+                    bmesh.ops.transform(bm, matrix=share_region_offset, space=share_region.matrix_world, verts=bm.verts)
+                    share_region.plasma_object.enabled = True
+                    share_region.hide_render = True
```
Hoikas, line 493:
```suggestion
                with utils.bmesh_object("{}_LinkingBook_ShareRgn".format(self.key_name)) as (share_region, bm):
                    bmesh.ops.create_cube(bm, size=(8.0))
                    share_region_offset = mathutils.Matrix.Translation(self.clickable.matrix_world.translation - share_region.matrix_world.translation)
                    bmesh.ops.transform(bm, matrix=share_region_offset, space=share_region.matrix_world, verts=bm.verts)
                    share_region.plasma_object.enabled = True
                    share_region.hide_render = True
                    yield share_region
```

Yielding the share_region is a secret incantation that tells Korman, "hey, this is an object we created temporarily for the export. Please delete it when we're done!"

korman/properties/modifiers/gui.py
```diff
@@ -474,7 +482,19 @@ def pre_export(self, exporter, bo):
         else:
             rgn_obj = self.clickable_region
 
-        yield self.convert_logic(bo, age_name=exporter.age_name, version=exporter.mgr.getVer(), region=rgn_obj)
+        # Generate the same six foot cube if we need a share region.
```
Hoikas, line 485: It's not quite six feet, and the comment should probably be closer to the cube generator ;)

korman/properties/modifiers/gui.py
```diff
@@ -474,7 +482,19 @@ def pre_export(self, exporter, bo):
...
+        if self.shareable:
+            if self.share_region is None:
+                with utils.bmesh_object("{}_LinkingBook_ShareRgn".format(self.key_name)) as (share_region, bm):
+                    bmesh.ops.create_cube(bm, size=(8.0))
+                    share_region_offset = mathutils.Matrix.Translation(self.clickable.matrix_world.translation - share_region.matrix_world.translation)
+                    bmesh.ops.transform(bm, matrix=share_region_offset, space=share_region.matrix_world, verts=bm.verts)
+                    share_region.plasma_object.enabled = True
+                    share_region.hide_render = True
+            else:
+                share_region = self.share_region
```
Hoikas, line 495:
```suggestion
            else:
                share_region = self.share_region
        else:
            share_region = None
```

This will make sure that `share_region` exists if the book is not sharable.

Hoikas, line 495:
```suggestion
            else:
                share_region = self.share_region
        else:
            share_region = None
```

This will make sure that `share_region` exists if the book is not sharable.

korman/properties/modifiers/gui.py
DoobesURU: I renamed it to `share_rgn` in the `_create_moul_nodes` def.

korman/properties/modifiers/gui.py
DoobesURU: Edited the comment and moved it just above the `create_cube` function.

korman/properties/modifiers/gui.py
DoobesURU: Indentation is everything! Fixed.

korman/properties/modifiers/gui.py
```diff
@@ -578,7 +601,7 @@ def _create_pots_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
         responder.link_output(responder_state, "state_refs", "resp")
         responder.link_output(linkingnode, "keyref", "respOneShot")
 
-    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region):
+    def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk_region, share_rgn):
```
Hoikas, line 604: I think I'd prefer for this one to stay `share_region` to match `clk_region` for consistency.

DoobesURU: Fair enough. This won't conflict with the share region node stuff below that way, will it?

Hoikas: You'll want to rename the share region node variable to something else

DoobesURU: Renamed to `share_msb_region` instead.

korman/ui/modifiers/gui.py
```diff
@@ -88,6 +88,9 @@ def row_alert(prop_name, **kwargs):
 
         layout.separator()
         layout.prop(modifier, "link_type")
+        if modifier.link_type == "kOriginalBook":
+            layout.prop(modifier, "shareable")
+            layout.prop(modifier, "share_region")
```
Hoikas, line 93: Seeing as how this only works on MOUL, I wonder if it should be moved down into the `if "pvMoul" in modifier.versions` section below. That way, it doesn't appear in TPotS only linking books.

DoobesURU: So you're thinking further down in the other `if "pvMoul"` among `link_destination` and `spawn_title` (presumably before or after both of those)?

Hoikas: Or you could change the `if` statement here to `if "pvMoul" in modifier.versions and modifier.link_type == "kOriginalBook"`. Whichever one makes the most sense to you in the UI.

DoobesURU: OK, think I got things looking the way I want. Thoughts are welcome.

Hoikas (approved):

PR #321, opened 2022-03-20
Add kDynamicBlocker collider toggle.

DoobesURU:
Exports a collider that should hopefully block only dynamic objects (kickables). Also, some formatting adjustments to the properties script to match others (this can be undone if wanted)

### DIFF
--- a/korman/exporter/physics.py
+++ b/korman/exporter/physics.py
@@ -165,28 +165,27 @@ class PhysicsConverter:
                 physical.friction = mod.friction
                 physical.restitution = mod.restitution
 
-                if not mod.dynamic_blocker:
-                    if mod.dynamic:
-                        if ver <= pvPots:
-                            physical.collideGroup = (1 << plSimDefs.kGroupDynamic) | \
-                                                    (1 << plSimDefs.kGroupStatic)
-                        physical.memberGroup = plSimDefs.kGroupDynamic
-                        physical.mass = mod.mass
-                        _set_phys_prop(plSimulationInterface.kStartInactive, simIface, physical,
-                                       value=mod.start_asleep)
-                    elif not mod.avatar_blocker:
-                        physical.memberGroup = plSimDefs.kGroupLOSOnly
-                    else:
-                        physical.memberGroup = plSimDefs.kGroupStatic
-
-                    # Line of Sight DB
-                    if mod.camera_blocker:
-                        physical.LOSDBs |= plSimDefs.kLOSDBCameraBlockers
-                        _set_phys_prop(plSimulationInterface.kCameraAvoidObject, simIface, physical)
-                    if mod.terrain:
-                        physical.LOSDBs |= plSimDefs.kLOSDBAvatarWalkable
-                else:
+                if mod.dynamic:
+                    if ver <= pvPots:
+                        physical.collideGroup = (1 << plSimDefs.kGroupDynamic) | \
+                                                (1 << plSimDefs.kGroupStatic)
+                    physical.memberGroup = plSimDefs.kGroupDynamic
+                    physical.mass = mod.mass
+                    _set_phys_prop(plSimulationInterface.kStartInactive, simIface, physical,
+                                   value=mod.start_asleep)
+                elif mod.dynamic_blocker and not mod.avatar_blocker:
                     physical.memberGroup = plSimDefs.kDynamicBlocker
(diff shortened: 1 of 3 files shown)


DoobesURU:
Currently, this is set up so that the collider object will only export with the kDynamicBlocker member group and will disable the other toggles, just to make them their own separate thing. This can be changed if that's too cautious.

korman/properties/modifiers/physics.py
```diff
@@ -69,18 +69,42 @@ class PlasmaCollider(PlasmaModifierProperties):
...
+
+    avatar_blocker = BoolProperty(name="Blocks Avatars",
+                                  description="Object blocks avatars",
+                                  default=True)
+    camera_blocker = BoolProperty(name="Blocks Camera LOS",
+                                  description="Object blocks camera line-of-sight",
+                                  default=True)
+    dynamic_blocker = BoolProperty(name="Blocks Dynamics",
+                                   description="Object blocks dynamic objects (kickables)",
+                                   default=False)
```
Hoikas, line 85: This would, by default, indicate that kickables should fall through the floor.

korman/exporter/physics.py
```diff
@@ -165,33 +165,36 @@ def generate_physical(self, bo, so, **kwargs):
...
-                        physical.collideGroup = (1 << plSimDefs.kGroupDynamic) | \
-                                                (1 << plSimDefs.kGroupStatic)
-                    physical.memberGroup = plSimDefs.kGroupDynamic
-                    physical.mass = mod.mass
-                    _set_phys_prop(plSimulationInterface.kStartInactive, simIface, physical,
-                                   value=mod.start_asleep)
-                elif not mod.avatar_blocker:
-                    physical.memberGroup = plSimDefs.kGroupLOSOnly
+                if not mod.dynamic_blocker:
+                    if mod.dynamic:
```
Hoikas, line 169: This combination indicates to me that kickables cannot collide with each other.

Hoikas (commented):
I would prefer to see these as additive - as in you specify which types of collision you want. I think that making any of these exclude one another would be confusing.

korman/properties/modifiers/physics.py
DoobesURU: > I would prefer to see these as additive - as in you specify which types of collision you want. I think that making any of these exclude one another would be confusing.

OK. I kinda figured that'd be overly-cautious. 😉

korman/exporter/physics.py
DoobesURU: > This combination indicates to me that kickables cannot collide with each other.

I changed that section of the code back to how it was while adding and modifying a new line that'll hopefully do the trick.

DoobesURU:
So I put the code into my copy of Korman and found a couple of things:

* It's actually supposed to be `kGroupDynamicBlocker` (see https://github.com/H-uru/libhsplasma/blob/master/core/PRP/Physics/plODEPhysical.h)
* I got the following error when trying to export:
`AttributeError: type object 'PyHSPlasma.plSimDefs' has no attribute 'kGroupDynamicBlocker'`

Maybe my libHSPlasma files are outdated?

dpogue:
> Maybe my libHSPlasma files are outdated?

No... So the issue is that libHSPlasma has to try to find common ground between 3 different physics engines (Havok, ODE, and PhysX). It looks like ODE and PhysX have a `kGroupDynamicBlocker`, but it's unclear if Havok does so there's no equivalent value in the (shared) plSimDefs.

There's some hacky logic to try to convert the plGenericPhysical data to the right per-engine flags. I _think_ you want to set the physical's `collideGroup` to `plSimDefs.kGroupDynamic`

DoobesURU:
@dpogue 's suggestion worked! Setting the collideGroup and the memberGroup to kGroupDynamic created a collider that allowed my avatar through but blocked Eddie the beach ball.

Assuming the UI placement of the new toggle is acceptable, I'd say this one's ready to merge with your approval. 😄

korman/ui/modifiers/physics.py
```diff
@@ -20,6 +20,7 @@ def collision(modifier, layout, context):
 
     split = layout.split()
     col = split.column()
+    col.prop(modifier, "dynamic_blocker")
```
Hoikas, line 23: Can you move this down to below `camera_blocker`? My CDO insists that everything be alphabetized.

Hoikas (approved):

PR #323, opened 2022-04-03
Fix syntax error in sound modifiers

dpogue:


Deledrius (approved):

Hoikas (approved):

DoobesURU:
Ugh. We'll that's embarassing. Thanks for catching that, @dpogue !

PR #324, opened 2022-04-04
Fix #316.

### LINKED ISSUE #316: Wavesets don't appear to be utilizing alpha or passindex
Not entirely sure when this cropped up, but it's something I've noticed in Veelay previously and now in Noloben.

The waveset, despite having a higher passindex and set for alpha, appears to still eat through objects with lower passindices and that also have alpha. Turning off the alpha on the affected objects fixes this, but makes properly blending textures difficult.

Is there a way to force the waveset to blend properly with alpha objects below it? Is there a combination of settings in the Blending mod that would work? So far, no combos in that have worked.

Hoikas:
@DoobesURU could you check this? I didn't bother because it seems like such a glaring omission...


DoobesURU:
Tried this on my Noloben and my waveset test blend. Waveset still eats through anything with alpha, even if it has a higher pass index.

Hoikas:
Closing per https://github.com/H-uru/korman/issues/316#issuecomment-1094094484

PR #335, opened 2022-06-25
Update to Linking Book Mod share function

DoobesURU:
Sets the default of the share function to false along with an extra check before creating a default share region. This ensures extra, unneeded share regions will not be added as is happening with the current setup (per Ehren's examination of the Chiso files).


Deledrius (approved):

korman/properties/modifiers/gui.py
```diff
@@ -482,7 +482,7 @@ def pre_export(self, exporter, bo):
         else:
             rgn_obj = self.clickable_region
 
-        if self.shareable:
+        if self.shareable and self.link_type == "kOriginalBook":
```
Hoikas, line 485: To reduce the amount of potentially duplicated logic, Korman usually has Python properties for this. Eg, in the class:

```python
    @property
    def export_share_region(self) -> bool:
        return self.shareable and self.link_type == "kOriginalBook"
```

Then this line becomes:
```suggestion
        if self.export_share_region:
```

korman/properties/modifiers/gui.py
```diff
@@ -435,6 +435,14 @@ class PlasmaLinkingBookModifier(PlasmaModifierProperties, PlasmaModifierLogicWiz
                                    description="Sets the share region in which the receiving avatar must stand",
                                    type=bpy.types.Object,
                                    poll=idprops.poll_mesh_objects)
+    drc_stamp = BoolProperty(name="DRC Stamped",
+                             description="Did the DRC stamp this book?",
```
Deledrius, line 439: Are Explorers making DRC-approved Ages now?

korman/properties/modifiers/gui.py
```diff
@@ -680,6 +688,16 @@ def _create_moul_nodes(self, clickable_object, nodes, linkingnode, age_name, clk
...
+        # Odds and Ends
+        stamped = nodes.new("PlasmaAttribBoolNode")
+        stamped.link_output(linkingnode, "pfm", "IsDRCStamped")
+        if not self.drc_stamp:
+            stamped.value = False
+        forcecam = nodes.new("PlasmaAttribBoolNode")
+        forcecam.link_output(linkingnode, "pfm", "ForceThirdPerson")
+        if self.third_person:
+            forcecam.value = True
+
```
Hoikas, line 700: I'm thinking we should defer these changes to another PR.

Deledrius: I agree.

PR #336, opened 2022-06-26
A batch of collider usability improvements.

Hoikas:
This addresses some common gotchas with exported colliders:
- Flat convex hulls (planes) are now forced to triangle meshes.
- Triangle mesh regions will cause the export to error on MOUL and Myst V mode because the physics engines in those games either do not support this combination or only partially support it.
- Triangle mesh regions now alert in the UI, where appropriate.

### DIFF
--- a/korman/exporter/physics.py
+++ b/korman/exporter/physics.py
@@ -264,6 +264,17 @@ class PhysicsConverter:
                 self._report.warn("{}: Physical memberGroup overwritten!", bo.name, indent=2)
                 physical.memberGroup = member_group
 
+        # Sanity checking: only TPotS/Havok fully supports triangle mesh detector regions.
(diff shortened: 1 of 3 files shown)


DoobesURU:
After exporting, it appears that it tried to abort when it found one bad object with collision:

```
Writing Age data...

Exported 'ChisoPreniv.age' in 99.22s

Aborted after 99.68s
Error: HdnDoorButton: Physical wants to be a convex hull but appears to be flat (volume=1.7214252068242786e-07), forcing to triangle mesh...
DAG zero... not allowed to happen!
Error: HdnDoorButton: Physical wants to be a convex hull but appears to be flat (volume=1.7214252068242786e-07), forcing to triangle mesh...
```

Hoikas:
Verified

PR #337, opened 2022-06-26
Fix #334.

### LINKED ISSUE #334: LOC elements reorder each session with Blender
While not really an issue on the user end, this is more for consideration for pull request/repo updates containing LOC files with multiple entries.

Each time an LOC file is exported, it jumbles up the various elements, which leads to unnecessary line changes each time the LOC is pushed to a new commit.

Could there be a way to order elements by when they were added, or perhaps more realistically, in alphabetical order based on element name?

EDIT: It's been determined that this reorganization only happens after quitting out of Blender, going back in, and re-exporting.

Hoikas:
This fixes #334 by sorting everything before writing it out to the localization file. Note that we need to sort each step along the way because the nested dicts are not sorted by a call to `sorted()` - it appears to only sort the key.

### DIFF
--- a/korman/exporter/locman.py
+++ b/korman/exporter/locman.py
@@ -150,7 +150,7 @@ class LocalizationConverter:
 
         def iter_element(element):
             if language_name is None:
-                yield from element.items()
+                yield from sorted(element.items())
             else:
                 yield language_name, element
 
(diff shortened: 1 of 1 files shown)


DoobesURU:
Unfortunately, it still seems to mix up the entries after opening Blender again.

DoobesURU:
With that latest force push, it looks like that worked! Elements are now being sorted alphabetically, even after closing and reopening Blender. Thumbs up on this fix!

PR #339, opened 2022-06-30
Fix #332.

### LINKED ISSUE #332: Blend textures break after multiple exports
One way to replace the AlphaBlend textures is to use the Blend type texture. This can also be used for flares and a few other odds and ends.

The problem is, after a few exports, the Blend textures stop exporting properly. Instead of the proper gradient, they export all black or white. Quitting out of Blender, restarting, and re-exporting fixes this, but the problem crops up again after multiple exports.

This is something I've noticed on a few Ages I've exported that use a few Blend textures (Direbo's islands being a big example) and I believe @Hazado recently encountered it when exporting my ripples and flares for a Relto page.

Hoikas:
Fix #332, I think.

CC @DoobesURU

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -1007,7 +1007,10 @@ class MaterialConverter:
                     value = func(x, y, width, height)
                     pixels[offset:offset+4] = (value,) * 4
             image = bpy.data.images.new(filename, width=width, height=height, alpha=True)
+            image.source = "GENERATED"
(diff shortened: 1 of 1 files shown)


Hoikas:
Verified.

DoobesURU:
After exporting Direbo, the Age that usually broke blends the most, eight times, I had no troubles. I believe this fix is good IMHO.

PR #341, opened 2022-07-01
Fix #331.

### LINKED ISSUE #331: Grass Shader copy modifier doesn't carry changed settings over
When copying the grass shader modifier from one object to another, the mod pastes to new objects with the default settings and does not carry over any of the changes made in the original mod. This happens with both copying the single mod and also copying it along with multiple mods to multiple objects.

Hoikas:
Fix #331. Pointer properties must be deep copied just like collections, except different.


Hoikas:
Verified.

PR #347, opened 2022-10-11
Add xSitCam.py functionality to Sitting Modifier

DoobesURU:
Adds new scripting to Sitting Modifier to use xSitCam.py if a camera is selected.

Note that, as of the first commit, the camera object does not attach to the proper node in the Python file script node and I'm not entirely sure why. **EDIT: This has since been fixed (see comment below).**

### DIFF
--- a/korman/properties/modifiers/avatar.py
+++ b/korman/properties/modifiers/avatar.py
@@ -122,10 +122,10 @@ class PlasmaSittingBehavior(idprops.IDPropObjectMixin, PlasmaModifierProperties,
                                  description="How far away we will tolerate the avatar facing the clickable",
                                  min=-180, max=180, default=45)
     sitting_camera = PointerProperty(name="Camera (optional)",
-                                  description="Activate a specific camera when sitting down",
-                                  type=bpy.types.Object,
-                                  poll=idprops.poll_camera_objects,
-                                  options=set())
+                                     description="Activate a specific camera when sitting down",
+                                     type=bpy.types.Object,
+                                     poll=idprops.poll_camera_objects,
+                                     options=set())
 
     def harvest_actors(self):
         if self.facing_enabled:
@@ -142,10 +142,12 @@ class PlasmaSittingBehavior(idprops.IDPropObjectMixin, PlasmaModifierProperties,
         if self.sitting_camera is not None:
             sittingpynode = self._create_python_file_node(tree, sitting_pfm["filename"], sitting_pfm["attribs"])
             sittingmod.link_output(sittingpynode, "satisfies", "sitAct")
+
             # Camera Object
             cameraobj = nodes.new("PlasmaAttribObjectNode")
             cameraobj.link_output(sittingnode, "pfm", "sitCam")
             cameraobj.target_object = self.sitting_camera
+
         # Clickable
         clickable = nodes.new("PlasmaClickableNode")
         clickable.link_output(sittingmod, "satisfies", "condition")
--- a/korman/properties/modifiers/avatar.py
+++ b/korman/properties/modifiers/avatar.py
@@ -79,7 +79,7 @@ class PlasmaLadderModifier(PlasmaModifierProperties):
 
 
 # Use xSitCam.py for when we want a camera to pop up
-sitting_pfms = {
+sitting_pfm = {
     "filename": "xSitCam.py",
     "attribs": (
         { 'id':  1, 'type': "ptAttribActivator", 'name': "sitAct" },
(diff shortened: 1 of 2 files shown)


Hoikas:
I like this idea. I'll read the code soon to see if there are any issues that are readily apparent.

DoobesURU:
Just found out why the Scene Object wasn't attaching to the script. Simple matter of capitalization (or lack thereof) in `ptAttribSceneobject`. Because of course it was. 😛 

From what I can tell, the new addition to the mod is now working properly and is ready for testing and potential merging!

korman/properties/modifiers/avatar.py
```diff
@@ -119,17 +134,62 @@ def harvest_actors(self):
...
+                facing.directional = True
+                facing.tolerance = self.facing_degrees
+            else:
+                # this socket must be explicitly disabled, otherwise it automatically generates a default
+                # facing target conditional for us. isn't that nice?
+                clickable.find_input_socket("facing").allow_simple = False
+        else:
+            sitting_pfm = sitting_pfms
+            sittingnode = self._create_python_file_node(tree, sitting_pfm["filename"], sitting_pfm["attribs"])
+            self._create_sitting_nodes(bo, tree.nodes, sittingnode)
```
Hoikas, line 172: It looks like this section is basically a duplicate of the old code, and the old code received a minor adjustment to implement the new functionality. I think it would be more straightforward to handle the two cases together. I'll provide a sample suggestion below, please let me know if I misunderstood something.

korman/properties/modifiers/avatar.py
```diff
@@ -119,17 +134,62 @@ def harvest_actors(self):
...
         clickable = nodes.new("PlasmaClickableNode")
         clickable.link_output(sittingmod, "satisfies", "condition")
         clickable.clickable_object = self.clickable_object
         clickable.bounds = find_modifier(self.clickable_object, "collision").bounds
 
+        # Camera Object
+        cameraobj = nodes.new("PlasmaAttribObjectNode")
+        cameraobj.link_output(sittingnode, "pfm", "sitCam")
+        cameraobj.target_object = self.sitting_cam
+
```
Hoikas, line 192:
```suggestion
        # Sitting Modifier
        sittingmod = nodes.new("PlasmaSittingBehaviorNode")
        sittingmod.approach = self.approach
        sittingmod.name = "SittingBeh"

        # xSitCam.py PythonFileMod
        if self.sitting_cam is not None:
            sittingpfm = self._create_python_file_node(tree, sitting_pfm["filename"], sitting_pfm["attribs"])
            sittingmod.link_output(sittingpfm, "satisfies", "sitAct")

        # Clickable
        clickable = nodes.new("PlasmaClickableNode")
        clickable.link_output(sittingmod, "satisfies", "condition")
        clickable.clickable_object = self.clickable_object
        clickable.bounds = find_modifier(self.clickable_object, "collision").bounds

        # Camera Object
        cameraobj = nodes.new("PlasmaAttribObjectNode")
        cameraobj.link_output(sittingnode, "pfm", "sitCam")
        cameraobj.target_object = self.sitting_cam

```

korman/properties/modifiers/avatar.py
```diff
@@ -78,6 +78,16 @@ def requires_actor(self):
         return True
 
 
+# Use xSitCam.py for when we want a camera to pop up
+sitting_pfms = {
```
Hoikas, line 82:
```suggestion
sitting_pfm = {
```

In conjunction with the other suggestion.

korman/properties/modifiers/avatar.py
```diff
@@ -111,6 +121,11 @@ class PlasmaSittingBehavior(idprops.IDPropObjectMixin, PlasmaModifierProperties,
     facing_degrees = IntProperty(name="Tolerance",
                                  description="How far away we will tolerate the avatar facing the clickable",
                                  min=-180, max=180, default=45)
+    sitting_cam = PointerProperty(name="Camera (optional)",
```
Hoikas, line 124: I think I would prefer for this to be `sitting_camera` instead.

korman/properties/modifiers/avatar.py
DoobesURU: Nah, I think I see what's going on. Nodes are created either way, and only if there's a camera is there a Python File node created and utilized. That certainly works (and looks) better.

korman/properties/modifiers/avatar.py
DoobesURU: Done!

DoobesURU:
Thanks for the feedback, @Hoikas ! I implemented your suggestions above. If there's others on the other three, I'll get at 'em after work.

korman/properties/modifiers/avatar.py
```diff
@@ -123,7 +138,14 @@ def logicwiz(self, bo, tree):
...
-
+        # xSitCam.py PythonFileMod
+        if self.sitting_camera is not None:
+            sittingpynode = self._create_python_file_node(tree, sitting_pfm["filename"], sitting_pfm["attribs"])
+            sittingmod.link_output(sittingpynode, "satisfies", "sitAct")
+            # Camera Object
+            cameraobj = nodes.new("PlasmaAttribObjectNode")
+            cameraobj.link_output(sittingnode, "pfm", "sitCam")
+            cameraobj.target_object = self.sitting_camera
         # Clickable
```
Hoikas, line 149: Adds a blank line

```suggestion

        # Clickable
```

PR #348, opened 2022-10-15
Add Imager Modifier

DoobesURU:
A potential Imager Modifier for Korman. Could eventually be used in a future Plasma premade object.

Everything appears to be working, but input on UI layout and any potential "gotchas" in properties I may have missed is certainly welcome.

### DIFF
--- a/korman/properties/modifiers/logic.py
+++ b/korman/properties/modifiers/logic.py
@@ -189,6 +189,11 @@ class PlasmaImager(PlasmaModifierProperties, PlasmaModifierLogicWiz):
                                      description="Texture slot used for the imager.",
                                      type=bpy.types.Texture,
                                      poll=_poll_texture)
+    imager_type = EnumProperty(name="Imager Type",
+                               description="Type of imager object will be.",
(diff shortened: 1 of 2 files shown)


DoobesURU:
Also not really sure if we want to expose some parts of the script such as the Inbox SDL variable, pellet score, members only, etc. or keep things more simple, so some feedback on that would also be appreciated.

DoobesURU:
Thinking I might also add a type of imager selector (postable or visitor) after the object, material, and texture are chosen to make the latter a bit more simple. For the latter though, would it be possible and prudent to add the extra Python scripting for a visitor imager to the Age's py file automatically or just leave that up to the user to do manually?

Hoikas:
I'm going to continue to defer on this because imagers in general are a bit fiddly on the Python side. You've exposed a lot of things, for example the visitor imager, but every Age that implements one of those has a set of copy-pasta'd Python code to actually update the imager. Ideally, we will rewrite xSimpleImager.py in the game to not be a steaming pile of 💩 and include the ability to update the imager as needed.

PR #349, opened 2022-10-17
Add Takable Clothing Clickable Mod

DoobesURU:
* Adds properties and UI for a potential clothing clickable mod that uses xTakableClothing.py

Note that this does *not* add the actual clothing item for the avatar to GlobalClothing, but will set up a clickable in an Age to utilize an existing one (or new one if it's been added otherwise).

### DIFF
--- a/korman/properties/modifiers/logic.py
+++ b/korman/properties/modifiers/logic.py
@@ -117,3 +117,163 @@ class PlasmaMaintainersMarker(PlasmaModifierProperties):
     @property
     def requires_actor(self):
         return True
+
+
+clothing_pfms = {
+    "filename": "xTakableClothing.py",
+    "attribs": (
+        { 'id':  1, 'type': "ptAttribString", 'name': "stringVarName" },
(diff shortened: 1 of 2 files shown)


DoobesURU:
Thoughts on a note in the mod mentioning the need to add any new clothing items to the GlobalClothing PRPs? Also, instead of the integers for colors, would a color picker be doable?

Deledrius:
> Also, instead of the integers for colors, would a color picker be doable?

It should be!  Who ever knows with the Blender API, but `template_color_picker` looks promising.

DoobesURU:
> > Also, instead of the integers for colors, would a color picker be doable?
> 
> It should be!  Who ever knows with the Blender API, but `template_color_picker` looks promising.
> 

I actually did try it. I had the UI looking good, but setting the RGB values separately into integers for the script kinda failed.

DoobesURU:
Switched the RGB integer values to color pickers to clean up the mod a bit. I think I got the equation right to convert the numbers from Blender to what URU uses, but probably needs to be double-checked.

PR #350, opened 2022-10-18
Add Yeesha/Relto Page Modifier

DoobesURU:
* Properties and UI for a simple modifier for `xYeeshaPages.py`.

This mod is for the clickable wiring only and assumes, if it's a new page, the modifications have been done to the aforementioned script and `GUI_District_YeeshaPageGUI.prp`.


Hoikas:
I think I'd prefer to defer until we're able to think carefully about the UX around the global Yeesha page GUI.

PR #352, opened 2022-11-16
Replace not-so-smart characters in journals.

Hoikas:
IIRC smart quotes at best don't display in journals. At worst they cause problems, so debounce them.


Hoikas:
@DoobesURU When you get a chance, could you verify that I didn't goof anything here?

DoobesURU:
> @DoobesURU When you get a chance, could you verify that I didn't goof anything here?

I'll check it out later today when I get out of work.

Hoikas:
Whoopsie daisy! That should be fixed now...

DoobesURU:
> Whoopsie daisy! That should be fixed now...

Yep, no error now. AFAICT, it seems to be working as it should!

PR #354, opened 2022-12-08
Telescope Modifier

DoobesURU:
Adds properties and UI for a modifier to set up a clickable telescope.

### DIFF
--- a/korman/properties/modifiers/logic.py
+++ b/korman/properties/modifiers/logic.py
@@ -142,7 +142,7 @@ class PlasmaTelescope(PlasmaModifierProperties, PlasmaModifierLogicWiz):
                                        description="Clickable object for telescope.",
                                        type=bpy.types.Object,
                                        poll=idprops.poll_mesh_objects)
-    region_object = PointerProperty(name="Click Region",
+    tele_region = PointerProperty(name="Click Region",
                                     description="Area where clickable becomes active.",
                                     type=bpy.types.Object,
                                     poll=idprops.poll_mesh_objects)
@@ -169,7 +169,7 @@ class PlasmaTelescope(PlasmaModifierProperties, PlasmaModifierLogicWiz):
 
         # Region
         telescoperegion = nodes.new("PlasmaClickableRegionNode")
-        telescoperegion.value = self.region_object
+        telescoperegion.region_object = self.tele_region
         telescoperegion.link_output(telescopeclick, "satisfies", "region")
 
         # Telescope Camera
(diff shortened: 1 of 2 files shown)


DoobesURU:
For some reason, the node tree is not being added. In the export log for my test run (BaronCityOffice), I found this:

```
'TelescopeNew01_DRAW' generated NodeTree 'TelescopeNew01_DRAW'
... 0 new object(s) were generated!
```

Any ideas where I may have goofed?

korman/properties/modifiers/logic.py
```diff
@@ -117,3 +117,122 @@ def export(self, exporter, bo, so):
...
+                                     poll=idprops.poll_empty_objects)
+    camera_object = PointerProperty(name="Camera",
+                                    description="Camera used when viewing through telescope.",
+                                    type=bpy.types.Object,
+                                    poll=idprops.poll_camera_objects)
+    telescope_name = StringProperty(name="Name (Optional)",
+                                    description="Name for your telescope (optional).",
+                                    options=set())
+
+    def LogicWiz(self, bo, tree):
```
Hoikas, line 161:
```suggestion
    def logicwiz(self, bo, tree):
```

korman/properties/modifiers/logic.py
```diff
@@ -117,3 +117,122 @@ def export(self, exporter, bo, so):
...
+                                     description="Empty object used to orientate avatar in front of telescope.",
+                                     type=bpy.types.Object,
+                                     poll=idprops.poll_empty_objects)
+    camera_object = PointerProperty(name="Camera",
+                                    description="Camera used when viewing through telescope.",
+                                    type=bpy.types.Object,
+                                    poll=idprops.poll_camera_objects)
+    telescope_name = StringProperty(name="Name (Optional)",
+                                    description="Name for your telescope (optional).",
+                                    options=set())
```
Hoikas, line 159: From what I can tell, this is the name of the GUI dialog to use for the telescope (rather than the name of the telescope itself, which is what the description above implies). Since we don't have the ability to export GUIs at the moment, I am tempted to say that this should be removed until we have some scaffolding around GUI dialogs, then we'll be able to make a more informed decision on how to expose this setting. For now, I think the value for `Vignette` in the PythonFileMod should be hardwired to `telescope`, which is the default global GUI for most telescopes in the game.

korman/properties/modifiers/logic.py
```diff
@@ -117,3 +117,122 @@ def export(self, exporter, bo, so):
...
+    clickable_object = PointerProperty(name="Clickable",
+                                       description="Clickable object for telescope.",
+                                       type=bpy.types.Object,
+                                       poll=idprops.poll_mesh_objects)
+    region_object = PointerProperty(name="Click Region",
+                                    description="Area where clickable becomes active.",
+                                    type=bpy.types.Object,
+                                    poll=idprops.poll_mesh_objects)
+    oneshot_object = PointerProperty(name="OneShot",
+                                     description="Empty object used to orientate avatar in front of telescope.",
```
Hoikas, line 150:
```suggestion
                                     description="Empty object used to orient avatar in front of telescope.",
```

Hoikas (commented):
When you get a chance, @Deledrius has some nice code that automatically generates the clickable and region objects that I think would be very helpful for this modifier. You can see examples of those on the ladder modifier.

korman/properties/modifiers/logic.py
Hazado: there are at least two different telescope options, could hardwire it for either

Hoikas: Can you clarify the name of the other option? I can't remember the name or what it looks like off the top of my head.

Hazado: me either, ill look into it with some prp snooping

Hazado: After looking through for a while, I cant find any others that use the xTelescope.py.
I found other telescope type stuff but they have custom scripts.
Everything uses `telescope` for xTelescope.py

DoobesURU:
Committed all the suggestions mentioned so far. Thanks for the help! Let me know if there's anything else you may notice.

DoobesURU:
Another batch of fixes. The modifier exports now and all seems to be connected properly in the Python mod...but the MSB animation states' wiring appears to be at the default node settings for all three and the animation fields are blank, so clearly I've done something wrong with the node settings there. Guessing my use of strings in that area is incorrect?

Hoikas:
It looks like you used `==` instead of `=` in many places where you wanted to set a value. For example,
```python
telescopename.value == "telescope"
```
should be
```python
telescopename.value = "telescope"
```

DoobesURU:
> It looks like you used `==` instead of `=` in many places where you wanted to set a value. For example,
> 
> ```python
> telescopename.value == "telescope"
> ```
> 
> should be
> 
> ```python
> telescopename.value = "telescope"
> ```

Heh, yeah. I noticed that before I saw this reply 😊 

Almost got this. Just need to figure out the facing mod and it should be exporting fine.

Hoikas (approved):

Issue #356, opened 2023-01-19
Support LoadMasks for graphics quality adjustments


dpogue:
In Cyan's Ages, some objects (particularly wavesets) are set up with plLoadMask flags to ensure that they are only rendered on the highest quality graphics level and when pixel shaders are present. Conversely, Cyan often provides a "fallback" water surface that is just a plane with some animated textures.

Currently, korman never exports plLoadMasks on any objects, which results in wavesets always rendering at any graphics quality level (or in some cases not rendering if pixel shaders aren't available). This has caused performance and memory issues for several people trying to visit fan Ages on lower spec machines or through tools like wine.

In PlasmaMax, these load masks are handled by a "Representation Set" which allows the artist to pick different objects to render at different quality levels, but it's not especially intuitive to use. It does have the benefit of allowing some sort of association between the objects shown at different levels.

The main alternative I see is to have a "Render Quality" modifier in korman that could be attached to individual objects to control the load mask for that object in particular. This would allow for finer-grained customization, but would require the artist to keep track of which objects are rendered at which quality levels.

---

plLoadMask should be set on a Scene Object, and all its owned keys (ObjInterfaces, Modifiers, etc.). It is part of the key used to refer to those objects (which is how the implementation actually works: It doesn't notify with a plRefMsg if it reads a key that shouldn't load under the current quality setting).

dpogue:
It appears that ShadowCasters and ShadowMasters also set load masks in PlasmaMax, based on the quality setting selected in the shadow properties.

PR #358, opened 2023-01-28
Add Quality modifier to provide load masks

dpogue:
This allows Age builder control over what quality levels objects are rendered at. This does not (yet) automatically hook up load masks for things like wavesets or grass shader mods, but provides a general quality modifier.

Screenshot from 2023-01-28 00-09-35

Related to #356.

### DIFF
--- a/korman/properties/modifiers/render.py
+++ b/korman/properties/modifiers/render.py
@@ -969,3 +969,59 @@ class PlasmaVisibilitySet(PlasmaModifierProperties):
             if not region.control_region:
                 raise ExportError("{}: Not all Visibility Controls are set up properly in Visibility Set".format(bo.name))
             addRegion(exporter.mgr.find_create_key(plVisRegion, bl=region.control_region))
+
+class PlasmaQualityLevels(bpy.types.PropertyGroup):
+    low = BoolProperty(name="Low Quality",
+                       description="Object renders at low quality graphics level",
+                       default=False,
+                       options=set())
+
+    mid = BoolProperty(name="Medium Quality",
+                       description="Object renders at medium quality graphics level",
+                       default=False,
+                       options=set())
+
+    high = BoolProperty(name="High Quality",
+                        description="Object renders at high quality graphics level",
+                        default=False,
+                        options=set())
+
+    ultra = BoolProperty(name="Ultra Quality",
+                         description="Object renders at ultra quality graphics level",
+                         default=False,
+                         options=set())
+
+    @property
+    def mask_value(self):
+        q = 0
+
+        if self.low:
+            q |= (1 << 0)
+
+        if self.mid:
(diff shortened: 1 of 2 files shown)


Hoikas:
Thanks for getting started on this. The load mask stuff is used in many, many places, so it would be very nice to see the two parts of the load mask (quality and capability) split out into different parts for us to use both in terms of the UI/properties and the exporting thereof. Also, I feel like one of those boxes should have checkmarks for the hardware's shader model level instead of quality markings.

dpogue:
> Also, I feel like one of those boxes should have checkmarks for the hardware's shader model level instead of quality markings.

So, it turns out plLoadMask doesn't actually support this. It's only "if any shaders are supported" or "if no shaders are supported". There are comments hinting that it was intended to be expanded to handle that case, but we can't (easily) do that now without breaking binary compatibility.

Rather than being a modifier, you'd rather this be always visible on every object, such as in the "Plasma object" panel?

Hoikas:
Hmmm... It sure looks like `plLoadMask::ComputeRepMasks()` is accepting a shader capability level 😕

dpogue:
Yep, but then it caps it to `kMaxCaps` which is 1: https://github.com/H-uru/Plasma/blob/63f457b952c51474c4359a6b4cbbae5ef2e373af/Sources/Plasma/CoreLib/plLoadMask.cpp#L167

See also: https://github.com/H-uru/Plasma/blob/63f457b952c51474c4359a6b4cbbae5ef2e373af/Sources/Plasma/CoreLib/plLoadMask.h#L65

Hoikas:
I've been thinking on an off about this a bit, and I think that, considering we really don't have fine grained control over what kind of hardware is being targeted, it doesn't really make sense to expose the dichotomy of whether or not the player's GPU supports shaders or not. All GPUs released in the last 15 years support shaders, so that's not a very useful yardstick. The reason why we care about shader support or not, from what I understand, is to ensure that things like wavesets and grass shaders don't display on systems that don't support them and instead show a fallback. This is, to my mind, an implementation detail, so we should probably hide that from the artist. I'm thinking that modifiers should tell Korman that they are exporting shaders, and modifiers that export shaders should accept an "alternative object(s)" parameter, which might be exposed here on the quality modifier. In this way, we can intelligently export the with/without shader bits without troubling the artist with a distinction that means nothing to them.

Deledrius:
I think that makes a lot of sense, and feels very Kormanic.

dpogue:
So, in that case, we'd just want 4 checkboxes for quality levels?

As for automatically adding shader load masks, I think GrassShaderMod is the easiest where the loadmask can just go on the modifier, and leave the rest of the object as-is for display. Wavesets will be more difficult and involve picking an alternative.

PR #362, opened 2023-02-15
Add Late Render Pass

Hazado:
Adds the Late export for blend pass exporting meshes

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -94,6 +94,8 @@ class _RenderLevel:
             return self._calc_level(self.MAJOR_FRAMEBUF)
         elif mods.test_property("draw_opaque"):
             return self._calc_level(self.MAJOR_OPAQUE)
+        elif mods.test_property("draw_late"):
+            return self._calc_level(self.MAJOR_LATE)
         elif mods.test_property("draw_no_defer"):
             blend_span = False
 
--- a/korman/properties/modifiers/render.py
+++ b/korman/properties/modifiers/render.py
@@ -50,7 +50,7 @@ class PlasmaBlendMod(PlasmaModifierProperties):
                                        ("OPAQUE", "Before Avatar", "Prefer for the object to draw before the avatar."),
                                        ("FRAMEBUF", "Frame Buffer", "Prefer for the object to draw after the avatar but before other blended objects."),
                                        ("BLEND", "Blended", "Prefer for the object to draw after most other geometry in the blended pass."),
-                                       ("LATE", "Late", "Prefer for the object to draw late in the blended pass.")],
+                                       ("LATE", "Late", "Prefer for the object to draw after all other alpha-blended objects.")],
                                 options=set())
     sort_faces = EnumProperty(name="Sort Faces",
                               description="",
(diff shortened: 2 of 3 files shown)


korman/properties/modifiers/render.py
```diff
@@ -49,7 +49,8 @@ class PlasmaBlendMod(PlasmaModifierProperties):
                                 items=[("AUTO", "(Auto)", "Let Korman decide when to render this object."),
                                        ("OPAQUE", "Before Avatar", "Prefer for the object to draw before the avatar."),
                                        ("FRAMEBUF", "Frame Buffer", "Prefer for the object to draw after the avatar but before other blended objects."),
-                                       ("BLEND", "Blended", "Prefer for the object to draw after most other geometry in the blended pass.")],
+                                       ("BLEND", "Blended", "Prefer for the object to draw after most other geometry in the blended pass."),
+                                       ("LATE", "Late", "Prefer for the object to draw late in the blended pass.")],
```
Deledrius, line 53: When is "late" in this context?  The other options are more explicit about when the object is drawn.

Hazado: it translates to an 8 for the major bit in the render pass

dpogue: "Late" happens after all other alpha-blended objects, and is being used (in this particular use case) to match the flags that Cyan uses on GZ Markers.

The full render-level behaviour: https://github.com/H-uru/Plasma/blob/3b2788c23f6437f2fd81a18a451173b73a92319a/Sources/Plasma/CoreLib/plRenderLevel.h#L49-L79

Deledrius: For the sake of the description, is that at the very end of the render pass, after the end, at the end of all render passes...?

Deledrius: "Prefer for the object to draw after all other alpha-blended objects." might be more descriptive then.

korman/exporter/mesh.py
```diff
@@ -94,6 +94,8 @@ def _determine_level(self, bo : bpy.types.Object, blend_span : bool) -> int:
             return self._calc_level(self.MAJOR_FRAMEBUF)
         elif mods.test_property("draw_opaque"):
             return self._calc_level(self.MAJOR_OPAQUE)
+        elif mods.test_property("draw_late"):
+            return self._calc_level(self.MAJOR_LATE)
```
Hoikas, line 98: This also needs to be applied to the blending modifier's dependencies at line 110.

korman/properties/modifiers/render.py
```diff
@@ -73,6 +74,10 @@ def draw_opaque(self):
     def draw_framebuf(self):
         return self.render_level == "FRAMEBUF"
 
+    @property
+    def draw_late(self):
+        return self.render_level == "LATE"
```
Hoikas, line 79: This property needs a stub on the `PlasmaModifierProperties` base class:
```python
    def draw_late(self):
        return False
```

to ensure that all modifiers will return a value for this property.

korman/exporter/mesh.py
Hazado: Im trying to understand the logic here, wont it return before it can ever get there?

Hoikas: If you have an object with the blending modifier that doesn't have a blend pass set but with render dependencies set, it may fall thorough to the case of checking the render dependencies below.

Hoikas (approved):

PR #365, opened 2023-03-08
Improve handling of `ptAttribNamedResponder`.

Hoikas:
When a responder node is linked to a Python file node's ptAttribNamedResponder socket, it will now export with the same name as the node itself. In that way, the responder can be addressed in the attribute's value mapping in the Python script itself.

Closes #346.

CC @DoobesURU for verification.


Hoikas:
@DoobesURU Does this meet your requirements?

Hoikas:
Pinging @DoobesURU for response

Hoikas:
Lack of feedback interpreted as implicit approval.

PR #366, opened 2023-03-18
Allow facing detectors for region sensors.

### LINKED ISSUE #360: SwimDockExit oneshot requires an "Object in Volume and Facing Detector"
Right now when the SwimDockExit one shot is setup, the region will be anything that enters the region makes the oneshot happen.
This is problematic for the swim exits because it will can cause the avatar that just entered the water from the exit area to trigger the oneshot and exit the water immediately when they are facing away from the water.

To be clear I am talking about the plObjectInVolumeAndFacingDetector and not a facing detector and object in volume region

Hoikas:
This allows Region Sensor nodes to export `plObjectInVolumeAndFacingDetector`, which only triggers the region sensor when an avatar is in the region, facing a certain direction (within a tolerance amount), and (optionally) moving forward.

BREAKING CHANGE: The Facing Target node can now only link to a single condition node.

A companion pull request is available for Plasma to ensure that swimming forward satisfies the moving forward condition. See H-uru/Plasma#1362.

A [test Age blend file](https://github.com/H-uru/korman/files/11008971/ObjectInVolumeAndFacing.zip) is available.

Closes #360.

### DIFF
--- a/korman/nodes/node_conditions.py
+++ b/korman/nodes/node_conditions.py
@@ -13,11 +13,14 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 import bpy
 from bpy.props import *
 from collections import OrderedDict
 import math
 from PyHSPlasma import *
+from typing import *
 
 from .node_core import *
 from ..properties.modifiers.physics import bounds_types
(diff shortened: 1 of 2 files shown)


Hoikas:
@Hazado Does this meet your requirements?

Hoikas:
Pinging @Hazado for response

Hoikas:
Lack of feedback interpreted as implicit approval.

PR #367, opened 2023-03-20
Simplify cube region generation code.

Hoikas:
This consolidates a very common bit of LogicWiz modifier boilerplate into one location for ease of use.

PR #370, opened 2023-04-04
Fix incorrect pretends-to-be-a semantics from #367.

Hoikas:
This fixes review comments left on #367 that were caused by missing pretends-to-be-a semantics. `BMeshObject` is intended to provide pass-through access to `bpy.types.Object`'s attributes.

Supercedes and closes #369.


Deledrius (approved):
This appears to address the issues I experienced. 👍🏼

PR #374, opened 2023-07-15
Fix CubeMap "No image" Bug

DoobesURU:
* Adds a sanity check that raises an error if a CubeMap doesn't have an image set to it.

### DIFF
--- a/korman/exporter/material.py
+++ b/korman/exporter/material.py
@@ -721,7 +721,7 @@ class MaterialConverter:
         # to a big "finalize" save step to prevent races. The texture cache would
         # prevent that as well, so we could theoretically slice-and-dice the single
         # image here... but... meh. Offloading taim.
-        self.export_prepared_image(texture=texture, owner=layer, indent=3,
+        self.export_prepared_image(texture=texture, owner=layer
                                    alpha_type=TextureAlpha.opaque, mipmap=True,
                                    allowed_formats={"DDS"}, is_cube_map=True, tag="cubemap")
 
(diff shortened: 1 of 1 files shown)


Hoikas (commented):
The diff would be smaller if you just did
```python
        if texture.image is None:
            raise ExportError(f"CubeMap '{texture.name}' has no cube image!")
```
and didn't put the already existing code into an `else:` block.

DoobesURU:
> ```python
> (f"CubeMap '{texture.name}' has no cube image!")
> ```

Fixed!

korman/exporter/material.py
```diff
@@ -719,7 +721,7 @@ def export_cubic_env(self, bo, layer, texture):
         # to a big "finalize" save step to prevent races. The texture cache would
         # prevent that as well, so we could theoretically slice-and-dice the single
         # image here... but... meh. Offloading taim.
-        self.export_prepared_image(texture=texture, owner=layer,
+        self.export_prepared_image(texture=texture, owner=layer, indent=3,
```
Hoikas, line 724:
```suggestion
        self.export_prepared_image(texture=texture, owner=layer
```

This seems to have slipped in

korman/exporter/material.py
```diff
@@ -719,7 +721,7 @@ def export_cubic_env(self, bo, layer, texture):
         # to a big "finalize" save step to prevent races. The texture cache would
         # prevent that as well, so we could theoretically slice-and-dice the single
         # image here... but... meh. Offloading taim.
-        self.export_prepared_image(texture=texture, owner=layer,
+        self.export_prepared_image(texture=texture, owner=layer
```
Hoikas, line 724:
```suggestion
        self.export_prepared_image(texture=texture, owner=layer,
```

PR #383, opened 2023-10-01
Enable export of "Auto Load" and "Local Only" page options.

Deledrius:


korman/exporter/convert.py
```diff
@@ -243,7 +243,8 @@ def _export_age_info(self):
         for page in age_info.pages:
             if page.enabled and ver in page.version:
                 external = page.page_type == "external"
-                mgr.create_page(age_name, page.name, page.seq_suffix, external=external)
+                flags = (plAgeInfo.kFlagPreventAutoLoad if page.auto_load else 0) | (plAgeInfo.kFlagIsLocalOnly if page.local_only else 0)
```
Hoikas, line 246: Yikes.... The idea here is that we pass the deets to `create_page` and let it do the flag fiddling. This line, though, is just not pretty. I suggest adding two more kwargs to `create_page`.

Hoikas (changes requested):
It looks like the changes that add the `auto_load` and`local_only` properties is missing.

Deledrius:
> It looks like the changes that add the `auto_load` and`local_only` properties is missing.

They already exist... and do nothing (as I found to my dismay when trying to use them): https://github.com/H-uru/korman/blob/master/korman/properties/prop_world.py#L125-L130

korman/exporter/convert.py
Deledrius: I didn't like it either, but I couldn't figure out a more pythonic way to bridge this gap (and there was an example of something similar elsewhere in the code) so I went with this.  I wasn't sure if you'd prefer more kwargs instead.

Deledrius: Updated: Cleaned this up and moved things to the other side of the function call.

Hoikas (approved):

PR #387, opened 2023-12-06
Fix #386.

### LINKED ISSUE #386: Add->Plasma->Linking Book broken
This problem seems to have cropped up with 0.15. When trying to add the Linking Book template from the Add->Plasma menu, this error comes up:

```
  File "D:\Korman 0.15\2.79\scripts\addons\korman\operators\op_mesh.py", line 559, in execute
    self.create_linkingbook_objects()
  File "D:\Korman 0.15\2.79\scripts\addons\korman\operators\op_mesh.py", line 590, in create_linkingbook_objects
    with utils.bmesh_object(clk_rgn_name) as (clk_rgn, bm):
AttributeError: module 'korman.exporter.utils' has no attribute 'bmesh_object'

location: <unknown location>:-1

location: <unknown location>:-1
```

My guess is that this happened when cube region generation was simplified in #367 , where `bmesh_object` was removed in favor of other coding.

Hoikas:
Fixes #386.

CC @DoobesURU

### DIFF
--- a/korman/operators/op_mesh.py
+++ b/korman/operators/op_mesh.py
@@ -587,7 +587,8 @@ class PlasmaAddLinkingBookMeshOperator(PlasmaMeshOperator, bpy.types.Operator):
         # Create Clickable Region
         clk_rgn_name = "{}_ClkRegion".format(self.name_stem)
         clk_rgn_size = 6.0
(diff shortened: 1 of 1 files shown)


DoobesURU:
Yep, just plugged that change into my script and the Linking Book setup is working again! :)

PR #392, opened 2023-12-26
Fix #391.

### LINKED ISSUE #391: GUI Dialog Mod missing parts
I just tried this with a fresh copy of 0.15. It seems that, while the Python mod and GUI Button Mod are created upon export, the Dialog Mod and Post Effects Mod are not for some reason. This was tested on three different Ages' GUI mods that had worked in earlier testing.

Hoikas:
GUI Dialog Mods don't require any particular object type - they're just hints about what's going on with the GUI. Limiting them to meshes actually breaks the Note Popup modifier, which attaches a GUI Dialog modifier to an empty.

CC @DoobesURU 

Fixes #391.


DoobesURU:
Perfect! I can confirm that this fix works on two regular Ages (Chiso and the Pub) and on the test Age blend I submitted in #391 . 😃

PR #395, opened 2024-01-13
Entry Cameras for the Spawn Point Modifier

Hoikas:
This simplies the creation of "entry cameras" as seen in Doobes' Ages Veelay and Chiso. In Doobes' Ages, the entry cameras are pushed by entering into a region around the spawn point. When exiting the region, the camera is popped off and the region is disabled. Unfortunately, since this is a general purpose modifier, we cannot simply use that approach. That approach assumes there is a single spawn point for the entire Age. If you were to enter a second spawn point, then walk intgo the other spawn point's entry camera region, it would be spuriously triggered. Therefore, I have written a new `xEntryCam.py` script for CWE to ensure that all entry cameras are disabled when the local player finishes linking in.

See H-uru/Plasma#1556

### DIFF
--- a/korman/exporter/utils.py
+++ b/korman/exporter/utils.py
@@ -20,6 +20,7 @@ import bpy
 import mathutils
 
 from contextlib import contextmanager
+import enum
 from typing import *
 
 from PyHSPlasma import *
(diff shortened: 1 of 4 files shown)


korman/exporter/utils.py
```diff
@@ -122,21 +123,53 @@ def create_camera_object(name: str) -> bpy.types.Object:
...
+    # Proxy over to the generic rectangular prisim implementation to
+    # ensure that it is well tested.
+    return create_box_region(name, (size, size, size), owner_object, origin)
+
+def create_box_region(
+        name: str, size: Union[Tuple[float, float, float], mathutils.Vector],
+        owner_object: bpy.types.Object,
+        origin: RegionOrigin = RegionOrigin.center
+) -> bpy.types.Object:
+    """Create a rectangular prisim region object"""
```
Deledrius, line 142:
```suggestion
    """Create a rectangular prism region object"""
```

Deledrius (changes requested):

Deledrius (approved):

Hoikas:
`xEntryCam.py` is now available in the offline KI and Plasma repos, so I'm merging this.

PR #396, opened 2024-01-15
Improve vertex color exporting significantly.

Hoikas:
This moves the piece-by-piece assembly of vertex colors from `_export_geometry()` into a one-stop-shop for getting the near final vertex colors as Blender knows them. Included in this is separating out the adjustment channels for wavesets - which are stuffed inside of vertex colors. It is now an error for a waveset to have a "col", "color", or "colour" vertex color layer. This is to prevent confusion. Wavesets now accept alpha (red), specularity (green), fresnel (blue), and edgelength (alpha) vertex color layers. The color values in these layers is averaged and output to the respective channels. Further, a default value for edgelength is now computed similar (but not exactly like) PlasmaMax's `SetWaterColor()` function. Artist input to the edgelength vertex color layer will modulate Korman's calculation.

CC @DoobesURU to verify nothing breaks and that waveset vertex alpha should now function as expected.

This is still a draft because I would like to do a pass over eliminating list comprehensions in the tight inner loop of `_export_geometry()`. Until Python 3.12, list comprehensions are actually implemented as function calls. I see 1 comprehension for every face and 1 comprehension for every vertex (!!!), which are major targets for performance optimization.

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -504,8 +504,13 @@ class MeshConverter(_MeshManager):
                     normal = source.normal if use_smooth else tessface.normal
 
                     # MOUL/DX9 craps its pants if any element of the normal is exactly 0.0
-                    normal = map(lambda x: max(x, 0.01) if x >= 0.0 else min(x, -0.01), normal)
-                    normal = hsVector3(*normal)
+                    # Profiling indicates that unrolling this would give a performance penalty
+                    # of roughly. Note: unrolling this calculation gives a 43% speedup.
+                    normal = hsVector3(
+                        max(normal[0], 0.01) if normal[0] >= 0.0 else min(normal[0], -0.01),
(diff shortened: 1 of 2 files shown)


DoobesURU:
At the moment, it appears to not only eat the textures, but we also have some sort of checker pattern on everything:
Desktop Screenshot 2024 01 15 - 18 58 18 26

Hoikas:
Can you share a screenshot with the results before this PR?

DoobesURU:
> Can you share a screenshot with the results before this PR?

Desktop Screenshot 2024 01 15 - 19 12 07 18
Should look like this.

Hoikas:
This PR is almost completely wrong and needs to be revised significantly.

Hoikas:
Closing in favor of rewriting the entire mesh converter, someday.

PR #399, opened 2024-01-18
Fix GUI popup's Post Effect FovX value

DoobesURU:
Small fix to adjust the FovX formula in the Post Effects Mod to make GUIs the proper width in-game.

### DIFF
--- a/korman/properties/modifiers/game_gui.py
+++ b/korman/properties/modifiers/game_gui.py
@@ -471,7 +471,7 @@ class PlasmaGameGuiDialogModifier(PlasmaModifierProperties, _GameGuiMixin):
 
         post_effect = exporter.mgr.find_create_object(plPostEffectMod, bl=bo)
         post_effect.defaultC2W, post_effect.defaultW2C = exporter.gui.convert_post_effect_matrices(camera_matrix)
-        post_effect.fovX = math.degrees(fov)
+        post_effect.fovX = math.degrees(fov * 1.15)
         post_effect.fovY = math.degrees(fov * (3.0 / 4.0))
         post_effect.hither = min((hither, yonder))
         post_effect.yon = max((hither, yonder))


DoobesURU:
So far, I have tested this fix on the several GUIs in Kirel and the few in BaronCityOffice and they all look correct with this new math.

dpogue (approved):
I tested this and it works, and seems to fix some horizontal compressing that happens with the default GUI cameras

Hoikas:
I'm really curious why this fudge factor is needed. In my testing, it's unnesecary, and the max plugin doesn't do anything like this IIRC. It would be nice to have a blend file that I could look at to see what's going on... I would hate for us to merge this fudge factor to fix one case but break another when we should really be fetching this value from somewhere else.

Jrius:
I don't know about MOUL, but at least for PotS this likely needs the same fix as regular cameras : [widescreen fix](https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/FeatureLib/pfCamera/plVirtualCamNeu.cpp#L274). Cyan messed up their FOVs everywhere.
In PotS GUIs usually all have the same low FOV so the distortion isn't as noticeable but it's always been there. Drizzle's widescreen mod fixes this, and it makes the KI a bit rounder from what I remember.

Hoikas:
I've been revisiting the "Game GUI" code recently - it was all added as an "experiment" with the implication that more would come but unfortunately never did. Anyway, I've been fleshing out the experiment, and I did notice some problems with the FOV. I remain unconvinced that the fix proposed here is correct... Mostly because I don't understand the significance of the `1.15` magic number. 

What I found in Korman 0.15 - 0.17:
- The "generate GUI camera" operator defaults to a FOV of 90 degrees. This FOV seems quite problematic in that the result in Blender does not match what is seen in Plasma. I'm not sure who is at fault here, but naievely multiplying the X FOV by 1.15 doesn't fix the problem. Changing the default to 45 degrees causes the behavior with operator generated cameras and automatic cameras to match. See 7aeb610b8c82181c85600df22445d014967f0f43
- The GUI scaling feature of the camera generator seemed to be broken in some situations due to a zero length normal returned by Blender. I'm not really sure why that is happening, but I added a mitigation for it. See 5edafc781b3537810ca751a14645c556dd57dc55

All that said, I'm going to close this PR. I do think there is still a problem with FOV, but I'm uncertain if it's with Korman or Plasma. If anyone figures it out and can explain the issue and the fix, that would be amazing!

PR #402, opened 2024-01-30
Lightmap AND Vertex Color Option

DoobesURU:
Adds an option in the drop-down menu to use both a lightmap _and_ a vertex color layer in the Bake Lighting modifier. This is used in certain instances such as the floor of the outer chamber of the private rooms in the hoods and Kirel.

This has worked in at least two instances, namely the aforementioned private room floor in Kirel and in the common library's interior top floor in Ae'gura.


DoobesURU:
Closed due to better solution with #404 .

PR #406, opened 2024-02-18
Debounce duplicate note popups

### LINKED ISSUE #405: Multiple clickables to one GUI PRP
So I'm not sure if this is a Korman issue or a Plasma issue (or both), but after having played with the Guild decorations in Kirel and setting them to an SDL Int/Byte variable, I added new clickables for each Guild GUI popup for their new positions in the "minimal" deco setting.

While doing this, I've noticed that having multiple clickables (in this case, two apiece) pointing to one GUI PRP causes the client to crash. I may also be getting this intermittently in Chiso as well for some odd reason.

Deactivating the duplicate clickables keeps the client from crashing when Linking in, FWIW.

Hoikas:
Fixes #405. Side effect: fixes potential error where generated objects crash when pre-exporting.

CC @DoobesURU

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -229,7 +229,7 @@ class Exporter:
                 for mod in bl_obj.plasma_modifiers.modifiers:
                     fn = getattr(mod, "sanity_check", None)
                     if fn is not None:
-                        fn()
+                        fn(self)
                 inc_progress()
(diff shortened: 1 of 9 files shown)


DoobesURU:
While it didn't seem to fix the problem I had in Chiso (who knows what's going on there), I _was_ able to get multiple clickables to point to the same GUI PRP without the client crashing this time in the test Kirel with Guild deco toggle, so the fix itself is sound.

PR #413, opened 2024-06-15
Use a more friendly error message for empty sound names.

Hoikas:


DoobesURU:
Worked exactly as advertised! I'd say this is good to go.

PR #416, opened 2024-07-08
GUIDialog Python Proc fixes

dpogue:
* Fixes an error when a PythonFileMod tries to set itself as the proc for a GUIDialog (which happens when a dialog is passed in as a Python attribute)
* Fixes a case where GUIDialogs would get added as pool objects to the scene node, but also (incorrectly) get attached as modifiers to a SceneObject. This can cause Uru to crash because things load in the wrong order (which is its own problem that I'm not spending more time debugging tonight)
* Allow creating node trees in GUI pages. Sometimes you want a PythonFileMod in a GUI page, or sometimes you want a ResponderModifier in a GUI page.
* Fix a case where all ATC anims were being listed as pool objects which is not necessary except in the case of avatar animations (and those should in future probably be special cased like GUI pages are)

### DIFF
--- a/korman/nodes/node_python.py
+++ b/korman/nodes/node_python.py
@@ -751,10 +751,10 @@ class PlasmaAttribObjectNode(idprops.IDPropObjectMixin, PlasmaAttribNodeBase, bp
         layout.prop(self, "target_object", text=self.attribute_name)
 
     def get_key(self, exporter, so):
-        attrib = self.to_socket
-        if attrib is None:
+        attrib_socket = self.to_socket
+        if attrib_socket is None:
             self.raise_error("must be connected to a Python File node!")
-        attrib = attrib.attribute_type
+        attrib = attrib_socket.attribute_type
 
         bo = self.target_object
         if bo is None:
@@ -788,7 +788,7 @@ class PlasmaAttribObjectNode(idprops.IDPropObjectMixin, PlasmaAttribNodeBase, bp
             if not gui_dialog.enabled:
                 self.raise_error(f"GUI Dialog modifier not enabled on '{self.object_name}'")
             dialog_mod = exporter.mgr.find_create_object(pfGUIDialogMod, so=ref_so, bl=bo)
-            dialog_mod.procReceiver = attrib.node.get_key(exporter, so)
+            dialog_mod.procReceiver = attrib_socket.node.get_key(exporter, so)
             return dialog_mod.key
 
     @classmethod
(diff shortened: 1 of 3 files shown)


korman/nodes/node_python.py
```diff
@@ -751,10 +751,10 @@ def draw_buttons(self, context, layout):
         layout.prop(self, "target_object", text=self.attribute_name)
 
     def get_key(self, exporter, so):
-        attrib = self.to_socket
-        if attrib is None:
+        pynode = self.to_socket
+        if pynode is None:
             self.raise_error("must be connected to a Python File node!")
-        attrib = attrib.attribute_type
+        attrib = pynode.attribute_type
```
Hoikas, line 757: I think your new variable name is a bit misleading.. We're getting the attribute's socket on the python file node, not the python file node itself. `attrib_socket` would probably be more appropriate.

korman/nodes/node_python.py
```diff
@@ -751,10 +751,10 @@ def draw_buttons(self, context, layout):
         layout.prop(self, "target_object", text=self.attribute_name)
 
     def get_key(self, exporter, so):
-        attrib = self.to_socket
-        if attrib is None:
+        pynode = self.to_socket
+        if pynode is None:
```
Hoikas, line 755:
```suggestion
        attrib_socket = self.to_socket
        if attrib_socket is None:
```

korman/nodes/node_python.py
Hoikas, line 757:
```suggestion
        attrib = attrib_socket.attribute_type
```

korman/nodes/node_python.py
```diff
@@ -787,8 +787,8 @@ def get_key(self, exporter, so):
             gui_dialog = bo.plasma_modifiers.gui_dialog
             if not gui_dialog.enabled:
                 self.raise_error(f"GUI Dialog modifier not enabled on '{self.object_name}'")
-            dialog_mod = exporter.mgr.find_create_object(pfGUIDialogMod, so=ref_so, bl=bo)
-            dialog_mod.procReceiver = attrib.node.get_key(exporter, so)
+            dialog_mod = exporter.mgr.find_create_object(pfGUIDialogMod, bl=bo)
+            dialog_mod.procReceiver = pynode.node.get_key(exporter, so)
```
Hoikas, line 791:
```suggestion
            dialog_mod = exporter.mgr.find_create_object(pfGUIDialogMod, so=ref_so, bl=bo)
            dialog_mod.procReceiver = attrib_socket.node.get_key(exporter, so)
```

Your pool object commit should prevent the `pfGUIDialogMod` from getting added to the modifiers list on the `ref_so`. Passing in `ref_so` will allow us to skip a dictionary lookup for the GUI page's location.

Hoikas (approved):

PR #417, opened 2024-08-23
Add Color Picker to DynText Modifier

DoobesURU:
Small addition to the Localized/DynText modifier to add a color picker for the font color in the font options. Utilizes the existing `font_color` def in properties.


DoobesURU:
Just realized export isn't working just yet, so converted to draft until I do. 😊

DoobesURU:
Looks like it wasn't anything I did, but an issue with Korman itself: #418

Hoikas (approved):
Is this ready to go?

DoobesURU:
> Is this ready to go?

Should be! Just one line of code to add a color picker below the other two font options.

Hoikas (approved):

PR #420, opened 2024-08-28
Add GUI Page type to pl_page_types for Modifiers

DoobesURU:
I noticed this while working on getting face/swivel mods working for Empty objects (another issue). The modifier properties use `room` and `gui` in `pl_page_types` to determine which modifier can work in which type of page. I found this wasn't working completely due to a missing reference in the `base.py` script. This fix adds the reference and gets the missing modifiers working for GUIs, such as the Advanced mod for Python scripts.

### DIFF
--- a/korman/properties/modifiers/base.py
+++ b/korman/properties/modifiers/base.py
@@ -32,7 +32,7 @@ class PlasmaModifierProperties(bpy.types.PropertyGroup):
     @property
     def allowed(self) -> bool:
         """Returns if this modifier is allowed to be enabled on the owning Object"""
-        allowed_page_types = getattr(self, "pl_page_types", {"room"})
+        allowed_page_types = getattr(self, "pl_page_types", {"room", "gui"})
         allowed_object_types = getattr(self, "bl_object_types", set())
         page_name = self.id_data.plasma_object.page
         if not allowed_object_types or self.id_data.type in allowed_object_types:


DoobesURU:
Note that I purposely excluded "external" for obvious reasons. 😉

dpogue:
I'm not sure that swivel mods make sense on objects that are part of a GUI? What was your use case here?

DoobesURU:
> I'm not sure that swivel mods make sense on objects that are part of a GUI? What was your use case here?

No, sorry. As I said, that's a separate issue. I just happened to notice it while digging in the scripts for it. This is just a fix to use the existing modifier definitions in the associated scripts in `properties/modifiers/`.

Hoikas:
This doesn't look quite right to me... My thinking was that only a very narrow set of modifiers would be useful in a GUI page, so, by default, we would only make a modifier available to room pages. Modifiers need to opt-into being available for gui pages. If there's any modifier that should be available in a GUI page but isn't, we'll want to add `pl_page_types = {"room", "gui"}` to that modifier specifically.

DoobesURU:
> This doesn't look quite right to me... My thinking was that only a very narrow set of modifiers would be useful in a GUI page, so, by default, we would only make a modifier available to room pages. Modifiers need to opt-into being available for gui pages. If there's any modifier that should be available in a GUI page but isn't, we'll want to add `pl_page_types = {"room", "gui"}` to that modifier specifically.

That seems to be already implemented into each script in `properties/modifier` already. This, as far as I can tell, is telling the `base.py` script to use those existing definitions in the other scripts. /shrug

dpogue:
Adding the `"gui"` option to a specific modifier worked for me in https://github.com/H-uru/korman/pull/416

PR #421, opened 2024-08-28
Allow FaceMod/Swivel on Empty Objects

DoobesURU:
Certain animated objects, such as the flames for the Menorah in the city and the New Messengers Pub, need an empty that can swivel for both functions to work properly. If an animated object has both scale animation and swivel, it doesn't animate, for example. I noticed this while converting the city.

This PR will allow EMPTY objects to have swivel/FaceMod. See city/Ae'gura files and menorah dummy objects for examples of this.

### DIFF
--- a/korman/properties/modifiers/render.py
+++ b/korman/properties/modifiers/render.py
@@ -798,7 +798,7 @@ class PlasmaViewFaceMod(idprops.IDPropObjectMixin, PlasmaModifierProperties):
     bl_category = "Render"
     bl_label = "Swivel"
     bl_description = "Swivel object to face the camera, player, or another object"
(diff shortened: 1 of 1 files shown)


DoobesURU:
Currently, the mod appears in the drop-down list, but does not appear upon clicking it in the list. Not entirely sure why this is happening, so it'll be a draft until then.

DoobesURU:
So as a lesson to never code while drinking, the mod was there the whole time and I erroneously thought it was the wrong one...so this actually does work. 😊

Hoikas (approved):
Makes sense. Empties do have coordinate interfaces, so this is valid albeit somewhat advanced.

PR #422, opened 2024-08-29
Fix External Page Export

DoobesURU:
Small fix to stop pages set to External from triggering the "invalid page" error.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -213,7 +213,7 @@ class Exporter:
 
                 if (default_enabled and not page) or (page in pages_enabled):
                     self._objects.append(obj)
-                elif page not in all_pages or page in external_pages:
+                elif page not in all_pages or page not in external_pages:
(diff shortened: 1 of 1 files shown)


korman/exporter/convert.py
```diff
@@ -213,7 +213,7 @@ def _collect_objects(self):
 
                 if (default_enabled and not page) or (page in pages_enabled):
                     self._objects.append(obj)
-                elif page not in all_pages or page in external_pages:
+                elif page not in all_pages or page not in external_pages:
```
Hoikas, line 216: I think this is correct as-is. What this is supposed to be doing is checking that all of the Plasma Objects are in pages that exists in the Plasma Age panel. If not, it reports that object as being in an invalid page. Further, we check if the object is in an external page. If it is, that's also an error because external pages are, by definition, not supposed to be created or touched by Korman, so no object in the blend file should be in an external page.

PR #426, opened 2024-11-10
Allow Python File nodes to activate each other.

### LINKED ISSUE #424: Feature Request: Receiver socket for Python File Node
I'm not sure how widely used the `Receivers` part of the Python File Mods is used, but for the sake of completeness, I figured I'd request the feature.

So far, I've only seen it used in the `Neighborhood nb01` Python File Mod `PythonFileImager02`, with the receiver used being `PythHoodInfoImagerUpdater`. Presumably, it's for exactly what it describes and needed for a successful hood export.

Again, not sure if this is needed or an outdated relic, but hopefully, it's an easy addition that's worth discussing. Thoughts?

Hoikas:
Fixes #424.

This is kind of a weird feature. You're now able to link a Python File node's (new) "Satisfies" socket to the `ptAttribActivator` socket on another Python File node. The Python File node that is the satisfier does not actually activate anything, though. It's really a way for the satisfied Python File node to communicate back to the satisfier.

CC @DoobesURU

### DIFF
--- a/korman/nodes/node_python.py
+++ b/korman/nodes/node_python.py
@@ -13,12 +13,16 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 import bpy
 from bpy.props import *
 
 from collections.abc import Iterable
 from contextlib import contextmanager
 from pathlib import Path
+from typing import *
+
 from PyHSPlasma import *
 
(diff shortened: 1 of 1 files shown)


DoobesURU:
Just wanted to mention that I saw this fix and have implemented it. I went to export Neighborhood, but I still need to animate a few rail cameras, it seems. 😊 

Anyway, when I get unrelated bits working, I'll let you know how the export goes.

PR #434, opened 2025-01-09
Correctly export vertex normals

Jrius:
As it is now, Korman does not correctly export vertex normals as they appear in Blender.

Blender computes actual mesh normals (=what is used by the viewport for shading) according to (non exhaustive list):
- shared vertices between polygons
- polygon smoothness (`use_smooth`)
- edge sharpness (`use_edge_sharp`)
- mesh auto smooth+angle (mesh `use_auto_smooth` & `auto_smooth_angle`). Very useful for modelling machines with both smooth and sharp angles !
- potential custom split normals data (not sure how it works to be honest)

On export, Korman currently tries to reuse common vertices between polygons (which is good), but is a bit too eager and reuses vertices which are part of polygons with different normals. It also doesn't use the right property to get the actual normal that respects all of the above (edge sharpness, auto smooth angle, etc). There are some times where it kinda works and others where it fails, but I won't get into the details.

What this MR does is retrieve the true normal (again, what is used by Blender's viewport, more or less) and use that when looking for an existing vertex / exporting a new vertex.

Pros:
- Lighting effects that involve vertex normals now behave closer to the Blender counterparts. Off the top of my head, this means runtime lighting, specular highlights, and reflection cubemaps. Very important IMHO, I'm a sucker for properly split normals.

Cons:
- Slightly heavier DrawableSpan due to more vertices being generated.
- May hit the vertex number limit earlier, even for people who don't want or need normal-based effects.

Backwards compatibility: previous files will see improvements if re-exported, if they relied on any of the previously mentioned effects (RT light, specular, cubemaps). No effect on baked lighting. May cause big meshes that previously exported fine to hit the vertex limit and fail.

Example of properly exported normals:
MR_normals_image
Left: Plasma result of a cubemapped object. Middle: mesh normals, seen using Blender's matcap. Right: edit mode displaying sharp edges.

### DIFF
--- a/korman/exporter/mesh.py
+++ b/korman/exporter/mesh.py
@@ -394,9 +394,11 @@ class MeshConverter(_MeshManager):
                 continue
 
             face_verts = []
-            use_smooth = tessface.use_smooth
             dPosDu = hsVector3(0.0, 0.0, 0.0)
             dPosDv = hsVector3(0.0, 0.0, 0.0)
+
+            # Unpack normals
+            tessface_normals = [tuple(n) for n in tessface.split_normals]
 
             # Unpack the UV coordinates from each UV Texture layer
             # NOTE: Blender has no third (W) coordinate
@@ -446,6 +448,7 @@ class MeshConverter(_MeshManager):
 
             # Convert to per-material indices
             for j, vertex in enumerate(tessface.vertices):
+                vertex_normal = tessface_normals[j]
                 uvws = tuple([tuple(uvw[j]) for uvw in tessface_uvws])
 
                 # Calculate vertex colors.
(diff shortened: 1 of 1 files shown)


korman/exporter/mesh.py
```diff
@@ -394,10 +394,12 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
                 continue
 
             face_verts = []
-            use_smooth = tessface.use_smooth
             dPosDu = hsVector3(0.0, 0.0, 0.0)
             dPosDv = hsVector3(0.0, 0.0, 0.0)
 
+            # Unpack normals
+            tessface_normals = [tuple(n) for n in tessface.split_normals]
```
Hoikas, line 401:
```suggestion
            tessface_normals = tessface.split_normals
```

korman/exporter/mesh.py
```diff
@@ -446,6 +448,7 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
 
             # Convert to per-material indices
             for j, vertex in enumerate(tessface.vertices):
+                vertex_normal = tessface_normals[j]
```
Hoikas, line 451:
```suggestion
                vertex_normal = tuple(tessface_normals[j])
```

Hoikas (commented):
This looks like a good improvement. I have one minor concern with the list comprehension added on line 451, though. In Python 3.11 and lower, list comprehensions are implemented as Python function calls under the hood. These are very expensive, and this is a fairly tight loop. I'd like to avoid this, if possible. The suggestion I'm attaching is what I think would do that.

Jrius:
Good catch, I did not know about that. On a light scene it does shave off a few dozen milliseconds - not much but may be useful for iterating on bigger scenes.

Hoikas (approved):

PR #435, opened 2025-01-09
Fix waveset export

Jrius:
Fixes waveset wind direction export when a wind object is specified. Looks like it got lost in the last refactor...

### DIFF
--- a/korman/properties/modifiers/water.py
+++ b/korman/properties/modifiers/water.py
@@ -186,9 +186,6 @@ class PlasmaWaterModifier(idprops.IDPropMixin, PlasmaModifierProperties, bpy.typ
                                   description="Object whose Y axis represents the wind direction and whose Z axis represents the water height",
                                   type=bpy.types.Object,
                                   poll=idprops.poll_empty_objects)
-    wind_speed = FloatProperty(name="Wind Speed",
-                               description="Magnitude of the wind",
-                               default=1.0)
     envmap = PointerProperty(name="EnvMap",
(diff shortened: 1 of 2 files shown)


dpogue (commented):
I'm pretty sure the wind direction and speed are automatically calculated by Plasma based on the reference object (and any animations that affect it), but might not be a bad thing to set initial values.

Jrius:
I don't think so, the wind direction object only exists in Blender/Max. On export it's used to compute `windDir`, which is just a vector, but the object itself is not linked to the plWaveSet7.

My PR description wasn't great - the point is to fix the undefined `wind_dir` variable. Currently, Korman fails to export any waveset with a wind object defined. If the wind object is not set, then it just reverts to the old PyPRP default.

Hoikas:
Good catch! The export would error without `wind_dir` being set. However,
as @dpogue mentions, every frame, the wind data in the waveset object will
be reset with the value from the wind object (read: kRefObject)'s CI. This
will cause the speed field to have no effect if a reference object is used.
Therefore, we should not consider the speed value in this case. It appears
that the speed value is completely unused and should probably be removed,
tbh.

Jrius:
Ah, I was looking at the PRC for a typical waveset and saw no plKey to any wind/ref object, but it's just omitted if the key is null. So that's why I thought the wind object was export time only. I see what you mean now.

Yeah, then wind speed has no effect. I'll remove it. You can't even control it by scaling the ref object, because the wind direction is normalized. Sad.

Hoikas (approved):

PR #436, opened 2025-02-08
Armatures and skinned objects

Jrius:
Alright, this is a big one. I probably won't have a lot of time to implement suggestions until the RAD is over, so feel free to take your time reviewing. I also didn't check how optimized the code is (especially regarding Python generators), suggestions are welcome.

The first thing that needs to be mentioned: I decided to let Korman generate a LOT of temporary Blender objects during export. This may feel messy, but I've rewritten the code several times and AFAICT is still the best solution. Temporary objects should all get cleaned up nicely whatever happens.

Second, there are still two small problems, that I'll get around to later:
- Warning messages about "Animation generated no applicators" - harmless.
- Exported skinned objects don't automatically get runtime lighting. This can be forced using a Lighting Info modifier, but it might be worth making it automatic just like other animated objects.

---

**Armatures**
- Armature objects are exported as regular `SceneObjects`.
- Each bone of the armature is exported as two (!) new `SceneObjects`. Bone hierarchy is respected.
- The two `SceneObjects` are: the rest position of the bone, and the actual deform bone. For instance, the following bone hierarchy:
    - Armature:
        - Bone1
            - Bone2
        - Bone3
- ...Will be exported as:
    - Armature:
        - Bone1_REST
            - Bone1
                - Bone2_REST
                    - Bone2
        - Bone3_REST
            - Bone3
- Exporting two `SceneObjects` per bone avoids a whole slew of problems when exporting animations. Otherwise we'd be running in the same problem as an object having a `matrix_parent_inverse`.
- For various reasons, those two bone objects are actually temporary Blender Empty objects that are created at export-time. (I rewrote this several times, but in the end this felt like the best solution.) (On cleanup: those empties are deleted.)

**Animations**
- Armatures with an Animation modifier automatically get their Animation Group modifier toggled on during export. (On cleanup: modifier gets re-disabled if necessary.)
- Animated bones generate two Empty objects as previously mentioned - one of those empties gets to be animated:
    - The bone's animations are copied to a temporary Action assigned to the empty object. (Cleanup: the Action is deleted.)
    - An Animation modifier that matches the armature's own is added to the empty object. This Animation modifier is referenced in the armature's own Animation Group modifier. (On cleanup: the reference is removed.)
- This means one can easily play specific armature animations from nodes/responders, as Korman will reroute those messages to the armature's Animation Group / `plMsgForwarder`.
- Note that just like Cyan, we generate one `plATCAnim` per bone, per animation. Ideally we would group all bones in a single `plATCAnim`, but in my tests this doesn't work - seems it's only available to special objects like avatars.
- Animation modifiers now have two extra properties to bake animations.
    - Baking is required to get bone inverse kinematic (IK) to export.
    - It also helps fix some occasional animations issues: wrong tangents, missing keyframes on armature bones, etc. For this reason, it's also available to regular animations.
    - Baking is neither a silver bullet (can worsen animations), nor cheap (takes a long time to compute), so it is disabled by default. It's up to the user to know when to enable it.

**Rigged/deformed meshes**
- If the armature itself is not exported, the modifier/pose will be baked in the mesh and no expensive runtime deformation happens. (This is useful to make variations of the same mesh. For instance: trees with different branch shapes and sizes.)
- Otherwise, the armature modifier is disabled (so it doesn't get baked into the mesh), and weight/indices are exported in the mesh's vertices. (On cleanup: modifier is reenabled.)
    - Vertices that are not deformed by any bone get assigned to a special "null" bone that never moves. This is very commonly used in Kemo for trees, since their roots don't move.
    - If possible, less than 3 skin indices will be used per vertex for better performances.
- When generating the Drawable Spans from the collection of Geometry Spans, a few things need to happen.
    - Bone transforms get added to the DSpan, and bones get a Draw Interface referencing said bone.
    - Icicles get adjusted matrix indices so they know which bones they are using.
- Worth noting that Blender objects can have multiple Armature modifiers. This is accepted by Korman, but results are completely untested.
- Rigged meshes do NOT get a `CoordinateInterface`, ever. Since people usually parent the rigged mesh to the armature itself, Korman will simply ignore that relationship on export.
    - The reason for this is armature sharing. Armatures can be shared between rigs in Blender, but the way `DrawableSpans` work, this would require duplicating the bones for each object, so screw that.
    - There is rarely a good reason for the object to have its own coordinate system since it's supposed to move with the armature anyway... (except maybe floating-point precision ?)

**Other**
- Regular (non deformable) objects can now be parented to armature bones. This provides an easy way to use IK for machines and the like, without the overhead of deforming the vertices.

Backwards compatibility: completely new feature so no problem expected unless I messed up. It does fix a bug with `TemporaryCollectionItem` though.

And here is a [test file](https://www.dropbox.com/scl/fi/pr4eklxlrbgc8rl9udwhm/armature_testfile.blend?rlkey=n83rxpiesqljf1spc0dj8y9mb&st=n8ta659n&dl=1) so you can see it in action.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -560,9 +560,7 @@ class Exporter:
                 so = self.mgr.find_create_object(plSceneObject, bl=bo)
                 self._export_actor(so, bo)
                 # Bake all armature bones to empties - this will make it easier to export animations and such.
-                for temporary_object in self.armature.convert_armature_to_empties(bo):
-                    # This could be a Blender object, an action, a context manager... I have no idea, handle_temporary will figure it out !
-                    handle_temporary(temporary_object, bo)
+                self.armature.convert_armature_to_empties(bo, lambda obj: handle_temporary(obj, bo))
             for mod in bo.plasma_modifiers.modifiers:
                 proc = getattr(mod, "pre_export", None)
                 if proc is not None:
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -21,6 +21,7 @@ import functools
 import inspect
 from pathlib import Path
 from typing import *
+from mathutils import Matrix
 
 from ..helpers import TemporaryObject
 from ..korlib import ConsoleToggler
@@ -28,6 +29,7 @@ from ..korlib import ConsoleToggler
 from PyHSPlasma import *
 
 from .animation import AnimationConverter
+from .armature import ArmatureConverter
 from .camera import CameraConverter
 from .decal import DecalConverter
 from . import explosions
@@ -56,6 +58,7 @@ class Exporter:
         physics: PhysicsConverter = ...
         light: LightConverter = ...
         animation: AnimationConverter = ...
+        armature: ArmatureConverter = ...
         output: OutputFiles = ...
         camera: CameraConverter = ...
         image: ImageCache = ...
@@ -80,6 +83,7 @@ class Exporter:
             self.physics = PhysicsConverter(self)
             self.light = LightConverter(self)
             self.animation = AnimationConverter(self)
+            self.armature = ArmatureConverter(self)
             self.output = OutputFiles(self, self._op.filepath)
             self.camera = CameraConverter(self)
             self.image = ImageCache(self)
@@ -252,12 +256,20 @@ class Exporter:
 
     def _export_actor(self, so, bo):
         """Exports a Coordinate Interface if we need one"""
+        parent = bo.parent
+        parent_bone_name = bo.parent_bone
+        offset_matrix = None
+        if parent_bone_name:
+            # Object is parented to a bone, so use it instead.
+            parent_bone = parent.data.bones[parent_bone_name]
+            parent = bpy.context.scene.objects[ArmatureConverter.get_bone_name(parent, parent_bone)]
+            # ...Actually, it's parented to the bone's /tip/. So we need to offset the child...
+            offset_matrix = Matrix.Translation((0, 0, -parent_bone.length))
         if self.has_coordiface(bo):
-            self._export_coordinate_interface(so, bo)
+            self._export_coordinate_interface(so, bo, offset_matrix)
 
         # If this object has a parent, then we will need to go upstream and add ourselves to the
         # parent's CoordinateInterface... Because life just has to be backwards.
-        parent = bo.parent
         if parent is not None:
             if parent.plasma_object.enabled:
                 self.report.msg(f"Attaching to parent SceneObject '{parent.name}'")
@@ -268,18 +280,25 @@ class Exporter:
                                  The object may not appear in the correct location or animate properly.".format(
                                     bo.name, parent.name))
 
-    def _export_coordinate_interface(self, so, bl):
+    def _export_coordinate_interface(self, so, bl, matrix: Matrix = None):
         """Ensures that the SceneObject has a CoordinateInterface"""
         if so is None:
             so = self.mgr.find_create_object(plSceneObject, bl=bl)
+        ci = None
         if so.coord is None:
             ci_cls = bl.plasma_object.ci_type
             ci = self.mgr.add_object(ci_cls, bl=bl, so=so)
-
+        if ci is not None or matrix is not None:
+            # We just created the CI, or we have an extra transform that we may have previously skipped.
+            matrix_local = bl.matrix_local
+            matrix_world = bl.matrix_world
+            if matrix is not None:
+                matrix_local = matrix_local * matrix
+                matrix_world = matrix_world * matrix
             # Now we have the "fun" work of filling in the CI
-            ci.localToWorld = utils.matrix44(bl.matrix_world)
+            ci.localToWorld = utils.matrix44(matrix_world)
             ci.worldToLocal = ci.localToWorld.inverse()
-            ci.localToParent = utils.matrix44(bl.matrix_local)
+            ci.localToParent = utils.matrix44(matrix_local)
             ci.parentToLocal = ci.localToParent.inverse()
             return ci
         return so.coord.object
@@ -333,6 +352,9 @@ class Exporter:
                     with indent():
                         mod.export(self, bl_obj, sceneobject)
             inc_progress()
+
+    def _export_armature_blobj(self, so, bo):
+        pass
 
     def _export_camera_blobj(self, so, bo):
         # Hey, guess what? Blender's camera data is utter crap!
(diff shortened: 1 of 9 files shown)


korman/exporter/mesh.py
```diff
@@ -387,6 +499,26 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
+                for bone in armature.data.bones:
+                    all_bone_names[bone.name] = i
+                    i += 1
+            # This will map the group ID (used by Blender vertices) to the bone index exported to Plasma.
+            # Theoretically supports multiple armatures, except if the two armatures have the same bone names (because that would be REALLY asking for a lot here).
+            # If the bone is not found, we'll just map to the null bone.
+            group_id_to_bone_id = [all_bone_names.get(group.name, 0) for group in bo.vertex_groups]
+            # We will also need to know which bones deform the most vertices per material, for max/pen bones.
+            for gd in geodata.values():
+                gd.total_weight_by_bones = { j: 0.0 for j in range(i) }
```
Hoikas, line 519:
```suggestion
            bone_count = len(all_bone_names)
            for gd in geodata.values():
                gd.total_weight_by_bones = { i: 0.0 for i in range(bone_count) }
```

Could this be a list instead of a dict?

korman/exporter/mesh.py
```diff
@@ -484,6 +616,37 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
 
+                    if export_deform:
+                        # Get bone ID and weight from the vertex' "group" data.
+                        # While we're at it, sort it by weight, and filter groups that
+                        # have no bone assigned. Take only the first 3 bones.
+                        weights = sorted([ \
+                                (group_id_to_bone_id[group.group], group.weight) \
+                                for group in source.groups \
+                                if group.weight > 0 \
+                            ], key=lambda t: t[1], reverse=True)
```
Hoikas, line 627: The `\` characters aren't needed here.

korman/exporter/mesh.py
```diff
@@ -484,6 +616,37 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
+                        if len(weights) > 3 and not warned_extra_bone_weights:
+                            warned_extra_bone_weights = True
+                            self._report.warn(f"'{bo.name}': only three bones can deform a vertex at a time. Please use Weight Tools -> Limit Total at 3 to ensure deformation is consistent between Blender and Plasma.")
```
Hoikas, line 630: I wonder if there's a way we can do that programmatically? I seem to remember that PlasmaMax will pick the bones with the highest weights.

korman/exporter/mesh.py
```diff
@@ -535,6 +698,39 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
+                # MaxBoneIdx and PenBoneIdx: these are the indices of the two highest-weighted bones on the mesh.
+                # Plasma will use those to compute a bounding box for the deformed object at run-time,
+                # in order to clip them when outside the view frustum.
+                # (The new BB is computed by extending the base BB with two versions of itself transformed by the max/pen bones).
+                # See plDrawableSpans::IUpdateMatrixPaletteBoundsHack.
+                # This is... about as reliable as you can expect: it kinda works, it's not great. This will have to do for now.
+                # (If you ask me, I'd rip the entire thing out and just not clip anything, framerate be damned.)
+                # Note that max/pen bones are determined by how many vertices they deform, which is a bit different (and more efficient)
+                # than whatever the Max plugin does in plMAXVertexAccumulator::StuffMyData.
+                sorted_ids_by_weight = sorted(((weight, id) for id, weight in data.total_weight_by_bones.items()), reverse = True)
```
Hoikas, line 711:
```suggestion
                sorted_ids_by_weight = sorted(((weight, id) for id, weight in data.total_weight_by_bones.items()), reverse=True)
```

korman/exporter/mesh.py
```diff
@@ -535,6 +698,39 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
+                geospan.maxBoneIdx = sorted_ids_by_weight[0][1]
+                geospan.penBoneIdx = sorted_ids_by_weight[1][1]
+
+                # This is also a good time to specify how many bones per vertices we allow, for optimization purposes.
+                max_deform_bones = data.max_deform_bones
+                if max_deform_bones == 3:
+                    geospan.format |= plGeometrySpan.kSkin3Weights | plGeometrySpan.kSkinIndices
+                elif max_deform_bones == 2:
+                    geospan.format |= plGeometrySpan.kSkin2Weights | plGeometrySpan.kSkinIndices
+                else: # max_bones_per_vert == 1
```
Hoikas, line 723: Let's do this explicitly and error if there's an unexpected value.

korman/exporter/mesh.py
```diff
@@ -535,6 +698,39 @@ def _export_geometry(self, bo, mesh, materials, geospans, mat2span_LUT):
...
+                    geospan.format |= plGeometrySpan.kSkin1Weight
+                    if len(group_id_to_bone_id) > 1:
+                        geospan.format |= plGeometrySpan.kSkinIndices
+                    else:
+                        # No skin indices required... BUT! We have assigned some weight to the null bone on top of the only bone, so we need to fix that.
+                        for vtx in data.vertices:
+                            weight = vtx.weights[0]
+                            weight = 1 - weight
```
Hoikas, line 731:
```suggestion
                            weight = 1.0 - weight
```

korman/exporter/convert.py
```diff
@@ -522,6 +559,11 @@ def _(temporary, parent):
             return temporary
 
         def do_pre_export(bo):
+            if bo.type == "ARMATURE":
```
Hoikas, line 562: This feels a little out of place, but I haven't wrapped my head around what's going on well enough to say what should change.

korman/exporter/convert.py
```diff
@@ -522,6 +559,11 @@ def _(temporary, parent):
...
+                so = self.mgr.find_create_object(plSceneObject, bl=bo)
```
Hoikas, line 563: I realize we can't do everything in one go, but I am thinking toward the future where we can (hopefully) export avatar animations from Korman, and IIRC avatar animation PRPs don't have a scene object in them at all.

korman/exporter/convert.py
```diff
@@ -522,6 +559,11 @@ def _(temporary, parent):
...
+                self._export_actor(so, bo)
+                # Bake all armature bones to empties - this will make it easier to export animations and such.
+                self.armature.convert_armature_to_empties(bo, lambda obj: handle_temporary(obj, bo))
```
Hoikas, line 566: This is where we need to be generator aware, instead of passing down this lambda.

Hoikas (commented):
This is a lot for me to wrap my head around. I think my major concern here is that I'm not sure that I like how the temporary bone empties are being handled. I worked pretty hard to try to establish a paradigm where temporary objects are `yield`ed to the exporter for lifetime management, and I'd like to see that continued.

You may want to submit the bugfixes in a separate PR so we can get them merged quickly.

PR #437, opened 2025-02-24
Fix TemporaryCollectionItem reference bug

Jrius:
Taken from #436. Fixes a random bug with `TemporaryCollectionItem`, where temporary items don't get cleaned up after export due to collection reallocation. So far I think it's only been used to generate extra sounds at export time, so it flew under the radar, but it happens very often with armatures.

### DIFF
--- a/korman/helpers.py
+++ b/korman/helpers.py
@@ -62,12 +62,14 @@ class GoodNeighbor:
 @contextmanager
 def TemporaryCollectionItem(collection):
     item = collection.add()
+    # Blender may recreate the `item` instance as the collection grows and shrink...
+    # Assign it a unique name so we know which item to delete later on.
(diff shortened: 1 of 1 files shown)


korman/helpers.py
```diff
@@ -62,12 +62,14 @@ def __exit__(self, type, value, traceback):
 @contextmanager
 def TemporaryCollectionItem(collection):
     item = collection.add()
+    # Blender may recreate the `item` instance as the collection grows and shrink...
+    # Assign it a unique name so we know which item to delete later on.
+    name = item.name = str(uuid4())
```
Hoikas, line 67: I don't think the `uuid4()` function exists yet. Is there a missing import?

Hoikas (approved):

PR #441, opened 2025-09-14
Convert RT light color to sRGB

Jrius:
As described on discord the other day. Realtime light color must be converted to sRGB, otherwise they will be much darker than in Blender and suffer from hue shift/oversaturation.

This is still not perfect. Maybe because Blender does all light calculation in linear space, whereas Plasma likely doesn't - or maybe there's still something else that I have missed. But at least it's much closer now.

Animated light color may still be slightly incorrect if the interpolation is done in the wrong color space. Not much that can be done without adding more keyframes.

Blender:
image
Plasma with fix (notice the middle row is still not as bright as it should be):
image

Test file:
[LightTest.zip](https://github.com/user-attachments/files/22321286/LightTest.zip)

Ages exported with previous versions of Korman will likely see their realtime lighting be brighter, and manual adjustment will be necessary.

### DIFF
--- a/korman/exporter/animation.py
+++ b/korman/exporter/animation.py
@@ -164,12 +164,12 @@ class AnimationConverter:
             self._exporter().report.warn("Cannot animate Lamp color because neither Diffuse nor Specular are enabled")
             return None
 
-        # OK Specular is easy. We just toss out the color as a point3.
+        # Specular must be converted to sRGB space (gamma correction).
         def convert_specular_animation(color):
             if lamp.use_negative:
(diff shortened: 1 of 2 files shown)


Hoikas:
This looks good to me. For my reference, could you make an image similar to the one labelled "Plasma with the fix" that shows the current in-game behavior so we can have a good before/after comparison?

Jrius:
Yep, here it is.
Before.
image

After.
image

Hoikas:
Oh, yeah, that's a huge improvement.

PR #449, opened 2025-12-13
Add link event trigger node.

Hoikas:
This is a higher level use case of the xLinkEventTrigger.py file introduced in H-uru/Plasma#1821. RIght now, this is just an example. I still have some iteration to do here, and I would like to add some more high level logic nodes to improve the workflow where Python has to get involved.

Python File nodes are tricky because the attributes are inputs to the script but function logically as outputs. For example, responders appear as an input to a Python file node, but Python file nodes can trigger responders. The inverse is also true. In the case of `xLinkEventTrigger.py`, we see the same thing - normally an activator is an input/trigger for a Python file node, but, in this case, the Python script is triggering the activator so that the responder or other things connected under "satisfies" can proceed.

### DIFF
--- a/korman/nodes/node_conditions.py
+++ b/korman/nodes/node_conditions.py
@@ -404,6 +404,106 @@ class PlasmaFacingTargetSocket(PlasmaNodeSocketBase, bpy.types.NodeSocket):
         return (not self.is_linked and not self.is_output)
 
 
+class PlasmaLinkEventNode(PlasmaNodeBase, bpy.types.Node):
+    bl_category = "CONDITIONS"
+    bl_idname = "PlasmaLinkEventNode"
+    bl_label = "Link Event"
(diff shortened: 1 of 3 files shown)


Hoikas:
This is now ready for feedback. I will only leave this PR up for feedback for ~24 more hours.

image

This node is a high level wrapper around xLinkEventTrigger.py and allows firing off an arbitrary responder or basic responder. I removed the whole activator activator thunk thing because I learned that activator activators do NOT forward event data, so information about which avatar did the trigger is lost. That means you cannot play animations on specific avatars.

Additionally, I have decided to remove the other high level logic nodes from this PR because they do not relate to this particular node, and none of them use the thunk method I was tinkering with.

Deledrius (approved):

PR #458, opened 2026-07-15
Add SDL Logic Nodes

Hoikas:
# Intro
This PR adds a set of logic nodes that can be used to more easily track the flow of execution and information. Almost all of these nodes export PythonFileMods under the hood. The Python File Nodes are great in that we have them, but it gets challenging when they are used with Responders because the flow control is inverted in that case - the responder node is an input to the PFM node and script, but the script can trigger the responder, meaning it should logically be an output. These nodes make it so that responders are outputs. See an example tree:

image

This tree combines two different situations and should probably be separated in a real Age. The upper tree is the more complicated one. It shows a clickable that will only be possible to click when the `Rm02BtnEnabled` SDL variable is set to true. When the clickable is pressed, the `Rm02SpinTorus` SDL boolean is toggled. When `Rm02SpinTorus` is true, the animation attached to the `Rm02Torus` object plays. When the variable is false, the same animation stops. This is the proper way to design logic interactions in the game to synchronize among multiple clients correctly, and the flow is quite clear IMO.

The bottom tree is an exercise for you, the viewer, to figure out. Be advised, although the "SDL Clickable" node is shown in the image above and described in the text above has actually been removed from this PR due to some design concerns. I will open a separate PR with that node and describe the design concerns there soon.

I do not have a demo blend to share at this time, so here is a list of the nodes added and the Python files they use in commit order. I may make a few more minor adjustments to this to ensure that some in-node lists start populated, but the code is otherwise complete, IMO. This is the follow-up promised in #449, and I expect I will want to merge this one shortly after it. Feedback needs to be prompt.

# Nodes
## Logic > SDL Variable
This just holds an SDL variable's name and shows logic flow. It doesn't actually export anything. 
image

## Logic > Change SDL
Changes an SDL value when notified.
Files: xAgeSDLBoolToggle, xAgeSDLBoolSet, xAgeSDLIntChange
image

## Logic > SDL Boolean Trigger
Trigger a responder when an SDL bool changes
Files: xAgeSDLBoolRespond, xAgeSDLBoolAndRespond
image

## Conditions > SDL Boolean Condition
Gates the firing of some other condition behind an SDL variable. Doesn't enable/disable the attached condition.
xAgeSDLBoolCondResp
image

## Logic > SDL Boolean Gate
Takes >=2 SDL variables as inputs and performs either a logical AND, OR, XOR, or NAND on them, storing the result in an output variable.
Filenames: xAgeSDLBoolAndSet, xAgeSDLBoolGateSet
image

## Logic > SDL Value Map
Uses an in-node lookup table to set the output variable value from the input variable value.
xAgeSDLVarSet
image

## Logic > SDL Integer Trigger
Fires specific responder states based on an SDL int value.
xAgeSDLIntStateListResp
image

 ## Logic > SDL Value Trigger
This is like SDL Boolean Trigger, except for a set of specific integer values.
xAgeSDLIntStartStopResp
image

## Bonus Chatter
Also, in order to "convert" from being inside of a Responder to triggering the Change SDL node, the "Trigger MultiStage" message node has been renamed to just "Trigger". The "Trigger" node can trigger a MultStage or any condition... such as Change SDL. This will be a powerful way to set SDL variables at specific points during responders, eg on a oneshot marker ("DoorButtonTouch" anyone?).

### DIFF
--- a/korman/nodes/node_conditions.py
+++ b/korman/nodes/node_conditions.py
@@ -490,6 +490,72 @@ class PlasmaLinkEventNode(PlasmaNodeBase, bpy.types.Node):
         self._add_py_parameter(pfm, 101, plPythonParameter.kString, self.trigger_for)
         self._add_py_parameter(pfm, 102, plPythonParameter.kString, self.trigger_at)
 
+
+class PlasmaSDLBoolConditionNode(PlasmaNodeBase, bpy.types.Node):
+    bl_category = "CONDITIONS"
+    bl_idname = "PlasmaSDLBoolConditionNode"
+    bl_label = "SDL Boolean Condition"
+
+    input_sockets: dict[str, dict[str, Any]] = {
+        "condition": {
+            "text": "Condition",
+            "type": "PlasmaConditionSocket",
+        },
+        "variable": {
+            "text": "Depends on SDL",
+            "type": "PlasmaSDLTriggererSocket",
+        },
+    }
+
+    output_sockets: dict[str, dict[str, Any]] = {
+        "satisfies_true": {
+            "text": "Trigger on True",
+            "type": "PlasmaConditionSocket",
+            "valid_link_nodes": {"PlasmaResponderNode", "PlasmaBasicResponderNode"}
+        },
+        "satisfies_false": {
+            "text": "Trigger on False",
+            "type": "PlasmaConditionSocket",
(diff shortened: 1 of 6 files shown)


Hoikas:
Rebased due to conflict from merging #449.

PR #460, opened 2026-07-20
SDL Clickable Node (Alternative)

Hoikas:
image

This is an alternate to the SDL Clickable node proposed in #459. This places the SDL variable as an input to the clickable node itself. In the aforementioned PR, the SDL variable and settings were slotted between the clickable and the Responder or PFM to "take ownership" of the clickable to ensure no one else tried to use it and be surprised when the hotspot was disabled. I then realized this could be short-circuited by simply making a different node tree and using the same object as a clickable again.

That led to the question of - well, what really is a re-used clickable? Do they have any value? Based on my understanding - no, they have no value. In practice `plSceneInputInterface` will only operate on *one* LogicMod attached to each SceneObject. So having multiple clickables or click-drags attached to a single object is nonsense. Therefore, this PR introduces a generic exclusive claim system such that a single object can only ever have one clickable associated with it. This is potentially a breaking change because nothing was stopping an artist from having the same object in multiple clickable nodes, but those don't make sense anyway. Adding this limitation will make things less surprising.

I think this approach is a lot clearer, but there is a danger that artists might start trying to use SDL to turn off their clickables while, for example, a OneShot is playing. That would be non-ideal because the clickable could easily be left in a basically permanently off situation if the player performing the OneShot crashes or force closes the game client during the OneShot.

I am inclined to merge this and close #459, but I will leave this open for feedback.

### DIFF
--- a/korman/exporter/convert.py
+++ b/korman/exporter/convert.py
@@ -13,6 +13,8 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Korman.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 import bpy
 
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


dpogue:
This definitely looks a bit clearer visually than the screenshots in #458

Issue #429, opened 2024-11-26
Option to cap a texture export to a maximum size


dpogue:
Thinking about this in the context of the work to get book pages consistent, and it would be nice to have a way to say "Use this 1024×1024 texture, but export it at a maximum size of 256×256" so that artists don't have to resize textures manually.

Based on `#1`, `#2`, etc. suffixes in Cyan-exported textures, I'm pretty sure Max has a way to do this, but I don't know how that works.

In our case, we'd need to decide if that happens on the image level, or the layer level, and how we'd differentiate the same texture being exported with different sizes.

Hoikas:
We probably don't want the same texture being exported at multiple sizes. I feel like the only time you'd want that is if you're doing something really esoteric with per-page textures and pages being paged in/out to manage a texture budget. At this point, modern hardware has enough VRAM that the engine itself will probably hit file size limitations before we run out of VRAM, so I don't think we need to bother with that kind of situation. I do think that being able to use a huge texture in Blender and scale it down on export is useful, however.

PR #461, opened 2026-07-23
Image Compression Improvements

Hoikas:
This is something of a grab-bag of improvements to the way images are exported. The onus for this change is that a recent update to GoMePubNew caused some complains of crashing, which we traced back to too much VRAM being used. Many fan Ages export lightmaps that are quite huge. Currently, Korman exports lightmaps as PNGs or RLE JPEGs, which means lightmaps are uncompressed in VRAM. To improve the VRAM usage, any lightmap 256x256 or greater in resolution will be compressed as a single level DXT1 texture. This should greatly improve the VRAM efficiency of lightmaps at the cost of slightly more disk space being used. This can be overridden using a new "Compression" option on the lightmap modifier.

To enable this change, I needed to remove the association Korman currently has where the "MipMap" button on the Blender Texture indicates both "export mip levels" *and* "compress to DXT". The "MipMap" button now only indicates "export mip levels". Compression options are now exposed on the image datablock and on image libraries.

Considering that lightmaps often have smooth gradients, I added the ability for Korman to internally set the squish color fit method used for DXT. Lightmaps use the slow `kBlockQualityUltra` to hopefully not impact the quality too much. Normal textures have been bumped to `kBlockQualityHigh`, which is actually libsquish's default compressor. It's noticeably slower than libHSPlasma's default. When textures are exported after this is merged, any textures over 512x512 will be noticeably slower the first time they are exported. The good news is that the texture cache will hit from then on, making this a one time expense for a potentially higher in game quality. When Korman was written, I don't think PyHSPlasma allowed setting the block quality, so this is really just paying the piper.

To hopefully balance out the disk space concern, I have added the ability for Korman to actually compress JPEGs for images used in ImageLibMods. This is now the default. Previously, any "JPEG" exported by Korman was just an RLE bitmap. Now, we will use the same JPEG/RLE logic that Cyan uses. This does mean that "uncompressed" lightmaps (and uncompressed images in general) can no longer be RLE maps. Instead, they will simply be PNGs or uncompressed bitmaps.

After exporting a recent-ish version of GoMePubNew, I found that these changes decreases the total potential image memory requirement of the Age from 406 MiB to 250 MiB. Actual VRAM usage is lower for both of those because the number includes journal images, which are not loaded into VRAM.

Future work would be to implement something like #429, but this is a good start.

### DIFF
--- a/korman/exporter/image.py
+++ b/korman/exporter/image.py
@@ -48,6 +48,7 @@ class _EntryBits(enum.IntEnum):
     last_export = 6
     image_count = 7
     tag_string = 8
+    quality = 9
 
 
 class _CachedImage:
@@ -63,6 +64,7 @@ class _CachedImage:
         self.modify_time = None
         self.image_count = 1
         self.tag = None
+        self.quality = None
 
     def __str__(self):
         return self.name
@@ -75,7 +77,7 @@ class ImageCache:
         self._read_stream = hsFileStream()
         self._stream_handles = 0
 
-    def add_texture(self, texture, num_levels, export_size, compression, images):
+    def add_texture(self, texture, num_levels, export_size, compression, quality, images):
         image, tag = texture.image, texture.tag
         image_name = str(texture)
         key = (image_name, tag, compression)
@@ -96,6 +98,7 @@ class ImageCache:
         image.image_data = images
         image.image_count = len(images)
         image.tag = tag
+        image.quality = quality
         self._images[key] = image
 
     def _compact(self):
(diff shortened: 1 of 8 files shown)


Deledrius:
Sounds like a lot of smart changes to me.  Any noticeable drop in lightmap quality?

Deledrius (approved):

Hoikas:
I've really only done superficial glances at the PRPs as of right now, but the results look acceptable to me. I suspect that the real test will be letting the artists loose on this.

Deledrius:
Agreed.

